Issue #93, opened 2018-02-28
/minted_token.all


kartsims:
# Subject of the issue

Can't retrieve the full list of tokens.

# Your environment

I am on a fresh Vagrant install, prepared using the `goban` repo.

# Steps to reproduce

- Create a fresh environment (I used `mix seed --sample`) and launch the server
- Create a new token using `/minted-token.create`
- Save the token ID 
- Call `/minted_token.get` on that token ID -> the token data is displayed
- Call `/minted-token.all` -> no data

This is the full JSON response I get from `.all` endpoint :

```
{
  "version": "1",
  "success": true,
  "data": {
    "pagination": {
      "per_page": 10,
      "is_last_page": true,
      "is_first_page": true,
      "current_page": 1
    },
    "object": "list",
    "data": []
  }
}
```

And this is the response I got from `.get` endpoint :

```
{
  "version": "1",
  "success": true,
  "data": {
    "updated_at": "2018-02-28T23:10:13.527129Z",
    "symbol": "SIMSIM",
    "subunit_to_unit": 100,
    "object": "minted_token",
    "name": "Simcoinzzz",
    "id": "SIMSIM:330acf3a-f46a-4dd8-b642-4c246d02da26",
    "created_at": "2018-02-28T23:10:13.527117Z"
  }
}
```

All tests have been done through Swagger only.

# Expected behavior

The token list should not be empty.

# Actual behavior

No token is available in that list.

kartsims:
The `mix seed` output Markdown is actually very different from the output of `mix seed --sample` so I figured I should try it out otherwise. 

Running `mix seed`afterwards and using that new seed didn't work.

Running `mix reset` to start off fresh (and it included a `mix seed`) didn't work either. I could still create the token, access it based on its ID, but still no luck on `/minted_tokens.all`.

mederic-p:
Hello, can you send the json body that you used in the request?
Make sure that you don't have a `search_term` params that would filter the results.

kartsims:
Yes that was it, I had to empty the `search_term` parameter, thank you !

I'd like to share my opinion on this endpoint before closing this issue if you don't mind :

- I'd except `search_term` to be empty in the example payload. In most APIs I've used, this is either an optional parameter or defaults to an empty string
- There seems to be no maximum for `per_page`. I didn't check the code implementation but I guess there actually is a maximum, even a high one (this might actually be a completely separate issue)
[...]

PR #96, opened 2018-03-06
Add maximum `per_page` value

unnawut:
Issue/Task Number: N/A

# Overview

Thanks to a great suggestion by kartsims in #93, this PR adds a maximum limit to `per_page` input to prevent a request overloading the system by sending a very high `per_page` value.

# Changes

- Updated `EWallet.Web.Paginator` to support `max_per_page`
- Added a configurable value in `ewallet/config/config.exs` called `max_per_page`
- Set the default `max_per_page` to 100
- Updated `docs/setup/env.md` to include `REQUEST_MAX_PER_PAGE`

# Implementation Details

- If the given `per_page` is higher than the `max_per_page`, the `per_page` will automatically be set to the `max_per_page` value instead of returning an error.

# Usage

Try send a request with a very high `per_page` value, you should see the response returning only up to the defined `max_per_page`.

You should also see that the response's `pagination.per_page` return the value that is actually used for the query, not the requested value.

This is only applicable to endpoints that are already supporting `per_page` attribute, e.g. `/admin/api/transaction.all`.

# Impact

You can optionally set the limit via an environment variable `REQUEST_MAX_PER_PAGE`. But apart from that, no extra steps required.

### DIFF
--- a/apps/ewallet/lib/ewallet/web/paginator.ex
+++ b/apps/ewallet/lib/ewallet/web/paginator.ex
@@ -7,6 +7,7 @@ defmodule EWallet.Web.Paginator do
   alias EWalletDB.Repo
 
   @default_per_page 10
+  @default_max_per_page 100
 
   defstruct [
     data: [],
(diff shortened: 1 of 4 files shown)


apps/ewallet/lib/ewallet/web/paginator.ex
```diff
@@ -7,6 +7,7 @@ defmodule EWallet.Web.Paginator do
   alias EWalletDB.Repo
 
   @default_per_page 10
+  @default_max_per_page 100
```
T-Dnzt: Why did you decide to add this `@default_max_per_page` even though we already define a default value in the config? 🤔

apps/ewallet/lib/ewallet/web/paginator.ex
```diff
@@ -31,15 +32,26 @@ defmodule EWallet.Web.Paginator do
     {:error, :invalid_parameter, "`page` must be non-negative integer"}
   end
   def paginate_attrs(_, %{"per_page" => per_page}) when not is_integer(per_page) or per_page < 1 do
-    {:error, :invalid_parameter, "`per_page` must be a non-native, non-zero integer"}
```
T-Dnzt: `native` 😂

apps/ewallet/lib/ewallet/web/paginator.ex
unnawut: A lot of the libraries have defaults at the library level. I think this guarantees that even if the values are missing from the config, the system will still work, so I want to follow this pattern more. E.g:

https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/supervisor.ex#L111-L132

https://github.com/mschae/cors_plug/blob/master/lib/cors_plug.ex#L4-L16

https://github.com/getsentry/sentry-elixir/blob/master/lib/sentry/config.ex#L8-L15

apps/ewallet/lib/ewallet/web/paginator.ex
unnawut: 🤷‍♂️

unnawut:
Rebased with the latest `develop` branch

PR #100, opened 2018-03-07
T65 - Prevent transactions between identical addresses

T-Dnzt:
Issue/Task Number: T65

# Overview

Currently, it is possible to create transactions between the same two addresses. This PR adds a mechanism preventing that.

# Changes

- Raise a `SameAddressError` error if the same address is found in both the debits and credits.

# Implementation Details

1. Extract the addresses for debits and credits.
2. Do an intersect and check if anything common was found.
3. Raise the `SameAddressError` if that's the case, else let the flow continue. 

# Usage

-

# Impact

-

### DIFF
--- a/apps/local_ledger/lib/local_ledger/entry/validator.ex
+++ b/apps/local_ledger/lib/local_ledger/entry/validator.ex
@@ -3,7 +3,19 @@ defmodule LocalLedger.Entry.Validator do
   This module is used to validate that the total of debits minus the total of
   credits for an entry is equal to 0.
   """
-  alias LocalLedger.Errors.{InvalidAmountError, AmountIsZeroError}
+  alias LocalLedger.Errors.{InvalidAmountError, AmountIsZeroError, SameAddressError}
+
+  def validate_different_addresses({debits, credits} = attrs) do
+    debit_addresses = extract_addresses(debits)
+    credit_addresses = extract_addresses(credits)
+    identical_addresses = intersect(debit_addresses, credit_addresses)
+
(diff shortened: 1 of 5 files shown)


apps/local_ledger/lib/local_ledger/entry/validator.ex
```diff
@@ -28,4 +40,6 @@ defmodule LocalLedger.Entry.Validator do
   defp total(list) do
     Enum.reduce(list, 0, fn(attrs, acc) -> attrs["amount"] + acc end)
   end
+  defp extract_addresses(list), do: Enum.map(list, fn e -> e["address"] end)
+  defp intersect(a, b), do: a -- (a -- b)
```
unnawut: Oh I never used `--` before 👏

apps/local_ledger/test/local_ledger/entry_test.exs
```diff
@@ -138,6 +138,35 @@ defmodule LocalLedger.EntryTest do
       assert error.errors == [correlation_id: {"has already been taken", []}]
     end
 
+    test "returns a 'same address' error when the from/to addresses are identical" do
+      genesis()
+
+      res = Entry.insert(%{
+        "metadata" => %{},
+        "debits" => [%{
+          "address" => "mederic",
```
unnawut: Did you not try to avoid personal names? :D

T-Dnzt: Oops, that's from a long time ago, will refactor in a future PR :D

PR #104, opened 2018-03-09
T99 - Refactor serializers

T-Dnzt:
Issue/Task Number: T99

# Overview

Serializers were a mess, some of them were duplicated, etc. This PR fixes that by putting all non-specific serializers in the `eWallet` app.

# Changes

- Remove `JSON` folder
- Replace `to_json` with `serialize`
- Move all non-specific serializers to `ewallet`
- Move serializers tests where they belongs after restructuring

# Implementation Details

Not much details here, it's mostly re-organizing the serializers.

# Usage

Nothing should change.

# Impact

Nope.


apps/ewallet_api/test/ewallet_api/v1/views/self_view_test.exs
```diff
@@ -35,24 +36,9 @@ defmodule EWalletAPI.V1.SelfViewTest do
...
-            %{
-              object: "minted_token",
-              id: token2.friendly_id,
-              symbol: token2.symbol,
-              name: token2.name,
-              subunit_to_unit: token2.subunit_to_unit,
-              metadata: %{},
-              encrypted_metadata: %{}
-            }]
+            MintedTokenSerializer.serialize(token1),
```
unnawut: Oh great idea.

unnawut:
@T-Dnzt made a good point about renaming `to_json/1` to `serialize/1` that I felt needing to note this down here.

**Question:**
> Since we removed `.JSON` namespace as well, how should we add a new response format?

**Answer:**
>The serializers don’t serialize in JSON, they format the schemas in maps that are then converted to JSON in the views.
>
> We currently have `def render("balances.json", ...`, we could have `def render("balances.xml", ...` and add an Elixir library that turns a bunch of nested maps into XML

PR #105, opened 2018-03-09
Fix 500 error when assigning a user with invalid email format

unnawut:
Issue/Task Number: T44

# Overview

This PR fixes 500 error when `/admin/api/account.assign_user` is requested with an invalid email format.

# Changes

- Changed all error returns from an atom to `{:error, _}` for consistency in handling them
- Added a test that submits /admin/api/account.assign_user with an invalid email format

# Implementation Details

Instead of adding another `else` condition, I refactored so that the `else` cases should always return `{:error, _}`. This makes some statements a little more wordy but overall producing less conditions in the logic, which should be better.

# Usage

Try make an authenticated call to /admin/api/account.assign_user with an invalid email (e.g. an email without @ sign). The response should return `success: false` with the error code `user:invalid_email`.

# Impact

Usual deployment. No extra steps.


unnawut:
Specifically assigned @ripzery to review since he found this bug.

PR #111, opened 2018-03-15
Client's request embedding for transaction request consumption

unnawut:
Issue/Task Number: T71

# Overview

This PR allows client to specify which extra association data to embed with the response.

# Changes

- Update all serializers to serialize `%Ecto.Association.NotLoaded{}` to nil
- Add `@embeddable` and `@always_embed` to TransactionRequestConsumptionController
- Embed the specified associations before sending the response

# Implementation Details

There were 3 options explored to handle when an association is not embedded:

```ex
# 1: Send nil
%{
  "success" => true, 
  "version" => "1",
  "data" => %{
    "parent"  => "data"
    "user_id" => "a43c801f-cc67-4047-bfb3-5cee967551de",
    "user"    => nil
  }
}

# 2: Send "not_loaded"
%{
  "success" => true, 
  "version" => "1",
  "data" => %{
    "parent"  => "data"
    "user_id" => "a43c801f-cc67-4047-bfb3-5cee967551de",
    "user"    => "not_loaded"
  }
}

# 3: Remove key completely
%{
  "success" => true, 
  "version" => "1",
  "data" => %{
    "parent"  => "data"
    "user_id" => "a43c801f-cc67-4047-bfb3-5cee967551de",
  }
}
```

We decided to go for 1 because:

1. `nil` is easy and performant for the client to handle, unlike option 2 that the client needs to check for a specific string.
2. The returned field keys are always consistent, unlike option 3.
3. The trade-off is that the client needs to distinguish between a nil that's caused by the association not embedded, or the object does not have an association, but this should be relatively straightforward since for all embeds, there will also be an `embed_id` in the response.

# Usage

Make a valid call to `/transaction_request.consume` and/or `/me.consume_transaction_request`

# Impact

Usual deployment. No extra steps required.

### DIFF
--- a/apps/admin_api/lib/admin_api/v1/serializers/membership_serializer.ex
+++ b/apps/admin_api/lib/admin_api/v1/serializers/membership_serializer.ex
@@ -2,6 +2,7 @@ defmodule AdminAPI.V1.MembershipSerializer do
   @moduledoc """
   Serializes membership(s) into V1 response format.
   """
+  alias Ecto.Association.NotLoaded
   alias AdminAPI.V1.UserSerializer
   alias EWalletDB.User
 
@@ -17,4 +18,6 @@ defmodule AdminAPI.V1.MembershipSerializer do
     |> Map.put(:account_role, membership.role.name)
     |> Map.put(:status, User.get_status(membership.user))
   end
+  def serialize(%NotLoaded{}), do: nil
+  def serialize(nil), do: nil
 end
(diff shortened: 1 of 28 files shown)


unnawut:
The implementation for `response_preload/2` should be extracted into a reusable module, but I took too much time trying to do that (tried `defmacro` along with `@embeddable` and `@always_embed`, they weren't nice to me 😭). 

So I'll do this as a separate task in the future.

unnawut:
If anyone is thinking I'm working late on a Friday night... nope. Went away for dinner and just came back to submit this.

PR #120, opened 2018-03-20
Serve fallback message when assets are missing & API key injection

unnawut:
Issue/Task Number: T136

# Overview

This PR adds a fallback message in case the front-end assets have not been built, improves the tests so that they reliably tests cases both when assets can be found and when they're not found, and auto-injects the API key pair from the backend to the admin panel.

# Changes

- Added `:dist_path` config to specify where to find the compiled frontend assets
- Add a fallback so that the admin panel serves a descriptive message if it cannot serve the assets.
- Update tests so they reliably tests both when the assets can be found and when they can't.
- Automatically injects API key ID and API key into the frontend app.

# Implementation Details

- `:dist_path` config is added so `AdminPanel.PageController` does not rely on a hardcoded path. This is useful for changing the path to the test assets when testing.
- **API key injection:** We decided to automatically inject the API key ID and API key from the backend Elixir subapp directly into the frontend serving page because:
  - The `.env` file needs to be set after a seed, this means extra steps to setup a system: 1) compile the apps 2) run the seed 3) configure `.env` with the seed information 4) compile frontend assets, instead of just 1) and 2)
  - Having to compile frontend assets with the `.env` file also means that server images can be prepared only with the backend, and the frontend will need to wait until runtime, which in inefficient.
  - The API key ID & API key for the Admin Panel is originally intended to be a client identifier, rather than a strong authenticator, so that we can do some quick and easy management of API clients (e.g. rate throttling, logging, etc.). All secure operations need to be conducted with the user authentication anyway. So this is acceptable as a temporary workaround to get the admin panel up and running while we wait to settle the discussion on the use of API keys in admin panel.

# Usage

Run the server with `mix omg.server` then delete the `/apps/admin_panel/assets/priv/dist/` folder. Instead of throwing error it should respond with a descriptive message that it could not find the assets.

Or delete the `/apps/admin_panel/assets/priv/dist/` folder, then run `mix test`. Instead of tests failing they should past now.

# Impact

Usual deployment.

### DIFF
--- a/apps/admin_panel/assets/src/omisego/config.js
+++ b/apps/admin_panel/assets/src/omisego/config.js
@@ -1,5 +1,5 @@
 export const ADMIN_PANEL_BASE_DIR = "/admin";
 
 export const ADMIN_API_BASE_URL = "/admin/api";
-export const ADMIN_API_KEY_ID = process.env.ADMIN_API_KEY_ID;
-export const ADMIN_API_KEY = process.env.ADMIN_API_KEY;
+export const ADMIN_API_KEY_ID = window.admin_api_conf.api_key_id;
+export const ADMIN_API_KEY = window.admin_api_conf.api_key;
--- a/apps/admin_panel/assets/src/omisego/config.js
+++ b/apps/admin_panel/assets/src/omisego/config.js
@@ -1,5 +1,5 @@
 export const ADMIN_PANEL_BASE_DIR = "/admin";
 
 export const ADMIN_API_BASE_URL = "/admin/api";
-export const ADMIN_API_KEY_ID = window.admin_api_conf.api_key_id;
-export const ADMIN_API_KEY = window.admin_api_conf.api_key;
+export const ADMIN_API_KEY_ID = window.adminConfig.apiKeyId;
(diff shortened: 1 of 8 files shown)


unnawut:
The API key injection is dependent on this PR and it's still open, so I took this chance to add in the API key injection as well.

apps/admin_panel/assets/src/omisego/config.js
```diff
@@ -1,5 +1,5 @@
 export const ADMIN_PANEL_BASE_DIR = "/admin";
 
 export const ADMIN_API_BASE_URL = "/admin/api";
-export const ADMIN_API_KEY_ID = process.env.ADMIN_API_KEY_ID;
-export const ADMIN_API_KEY = process.env.ADMIN_API_KEY;
+export const ADMIN_API_KEY_ID = window.admin_api_conf.api_key_id;
```
sirn: 🥇 🥇 🥇 

However, JavaScript convention is to use `camelCase` for variable and `UNDERSCORE_NAME` only for constants, so this is probably better as `window.adminConfig.apiKeyId` or something.

PR #122, opened 2018-03-22
Fix alert box overlap the logo on the login page

ripzery:
Issue/Task Number: `154-fix-alert-overlap-the-logo`
 
# Overview

The alert box on the login page doesn't look nice because it overlaps the logo. This PR will solve it.

# Changes

- A better alert box

**From**
image

**To**
image

# Implementation Details

- Add `display: inline-block` to `.alert` class
- Create class `.alert-center`
- Apply `.alert-center` to the root of the `Alerter` component

# Usage

`N/A`

# Impact

`N/A`

### DIFF
--- a/apps/admin_panel/assets/src/components/Alerter.jsx
+++ b/apps/admin_panel/assets/src/components/Alerter.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const Alerter = ({ alert }) => (
-  <div>
+  <div className="alert-center">
     {alert.message ? (
       <div className={`alert ${alert.type}`}>
         {alert.message}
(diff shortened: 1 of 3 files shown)


apps/admin_panel/assets/src/components/Alerter.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const Alerter = ({ alert }) => (
-  <div>
+  <div className="alert-center">
```
unnawut: Not sure about the best practice but maybe move `alert-center` to stay with `alert alert-center ${alert.type}` to keep the markup clean?

ripzery: The result is different because the div area is different

image

PR #126, opened 2018-03-23
Implement account level limit

unnawut:
Issue/Task Number: T157

# Overview

To prevent unnecessary complexity without observing real-life use cases first, we have decided to limit the number of account levels to 1 (one master account with uncapped direct children).

# Changes

- Add a new function `Account.get_depth/1` that returns the depth of the given account relative to the master account.
- Add new validator `AccountValidator.validate_account_level/2` that checks that the inserting/updating account does not go beyond the given max account level.
- Add this new validator above to the Account's changeset.

# Implementation Details

N/A

# Usage

Call `/admin/api/account.create` or `/admin/api/account.update` and try to set the account's parent to a parent that already reached the maximum account level (i.e. you should not be able to create an account that does not have the master account as its parent).

# Impact

N/A

### DIFF
--- a/apps/admin_api/priv/spec.yaml
+++ b/apps/admin_api/priv/spec.yaml
@@ -1672,8 +1672,6 @@ components:
                 type: string
               description:
                 type: string
-              master:
-                type: string
               parent_id:
                 type: string
               metadata:
@@ -1685,7 +1683,6 @@ components:
             example:
               name: "Account Name"
               description: "The account description"
-              master: false
               parent_id: "ce3982f5-4a27-498d-a91b-7bb2e2a8d3d1"
               metadata: {}
               encrypted_metadata: {}
@@ -1703,8 +1700,6 @@ components:
                 type: string
               description:
                 type: string
-              master:
-                type: string
               metadata:
                 type: object
               encrypted_metadata:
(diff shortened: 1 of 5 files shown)


apps/admin_api/priv/spec.yaml
```diff
@@ -1683,7 +1685,8 @@ components:
             example:
               name: "Account Name"
               description: "The account description"
-              master: true
+              master: false
```
mederic-p: What happens if we put true here? 🤔

unnawut: Oh good question. Actually the master account is determined by the system now, there will always be a master account and the rest are non-master. This attribute is not valid anymore (for create/update, but is still available for get/all). Will remove.

mederic-p: Yep, that's what I was thinking about, but was not sure :)

apps/ewallet_db/lib/ewallet_db/account.ex
```diff
@@ -189,4 +197,42 @@ defmodule EWalletDB.Account do
     |> where([b], b.account_id == ^account.id)
     |> Repo.one()
   end
+
+  @doc """
+  Determine the relative depth to the master account.
+
+  The master account has a relative depth of 0.
+  """
+  def get_depth(%Account{} = account), do: get_depth(account.id)
```
T-Dnzt: Please add the `specs` to all the functions so we know the input / ouputs.

unnawut:
@T-Dnzt Please review again

We need another Ecto version bump to get the nice `Account.t`, `Balance.t` etc. for typespec. So for now I used `%Account{}` and `%Balance{}` which are also [allowed by the guides](https://elixir-lang.org/getting-started/typespecs-and-behaviours.html).

PR #128, opened 2018-03-27
Improve insufficient funds error

T-Dnzt:
Issue/Task Number: T9

# Overview

This PR uses the `subunit_to_unit` from minted token to present more readable errors when funds are not sufficient for a transaction.

# Changes

- Move error message to eWalletAPI / AdminAPI for insufficient funds
- Format amounts as floats
- Update tests

# Implementation Details

In order to show the actual amount, we need the minted token which cannot be retrieved in the local ledger, therefore it was moved in the API. 

# Note

The code is duplicated in the error handlers, but the refactoring of the error handling is coming soon and will unify those.

### DIFF
--- a/apps/admin_api/lib/admin_api/v1/error_handler.ex
+++ b/apps/admin_api/lib/admin_api/v1/error_handler.ex
@@ -140,8 +140,8 @@ defmodule AdminAPI.V1.ErrorHandler do
   Handles response of insufficient funds error.
   Duplicate from eWalletAPI ErrorHandler, will be refactored in error merge.
   """
-  def handle_error(conn, :insufficient_funds, d) do
-    handle_error(conn, "transaction:insufficient_funds", d)
+  def handle_error(conn, :insufficient_funds, data) do
+    handle_error(conn, "transaction:insufficient_funds", data)
   end
   def handle_error(conn, "transaction:insufficient_funds", %{
     "address" => address,
--- a/apps/admin_api/lib/admin_api/v1/error_handler.ex
+++ b/apps/admin_api/lib/admin_api/v1/error_handler.ex
@@ -7,6 +7,7 @@ defmodule AdminAPI.V1.ErrorHandler do
   import Plug.Conn, only: [halt: 1]
   alias Ecto.Changeset
   alias EWallet.Web.V1.{ErrorSerializer, ResponseSerializer}
+  alias EWalletDB.MintedToken
 
   @errors %{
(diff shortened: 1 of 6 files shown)


apps/admin_api/lib/admin_api/v1/error_handler.ex
```diff
@@ -130,10 +131,37 @@ defmodule AdminAPI.V1.ErrorHandler do
...
+    handle_error(conn, "transaction:insufficient_funds", d)
+  end
+  def handle_error(conn, "transaction:insufficient_funds", %{
+    "address" => address,
+    "current_amount" => current_amount,
+    "amount_to_debit" => amount_to_debit,
+    "friendly_id" => friendly_id
+  }) do
+    minted_token = MintedToken.get(friendly_id)
+    current_amount = :erlang.float_to_binary(current_amount / minted_token.subunit_to_unit,
```
unnawut: Let's create or use a `Money` module (as a separate task)?

apps/admin_api/lib/admin_api/v1/error_handler.ex
```diff
@@ -130,10 +131,37 @@ defmodule AdminAPI.V1.ErrorHandler do
...
+
   def handle_error(conn, :invalid_parameter, description) do
     respond(conn, @errors.invalid_parameter.code, description)
   end
 
+  @doc """
+  Handles response of insufficient funds error.
+  Duplicate from eWalletAPI ErrorHandler, will be refactored in error merge.
+  """
+  def handle_error(conn, :insufficient_funds, d) do
```
unnawut: `data` instead of `d` maybe?

apps/ewallet_api/lib/ewallet_api/v1/error_handler.ex
```diff
@@ -111,6 +112,31 @@ defmodule EWalletAPI.V1.ErrorHandler do
     respond(conn, code, description, messages)
   end
 
+  @doc """
+  Handles response of insufficient funds error.
+  """
+  def handle_error(conn, :insufficient_funds, d) do
```
unnawut: `data` instead of `d` maybe?

apps/admin_api/lib/admin_api/v1/error_handler.ex
T-Dnzt: Okay, we'll do it in a future task!

PR #134, opened 2018-03-29
External IDs for Account schema

unnawut:
Issue/Task Number: T147

# Overview

This PR adds the ability to add and autogenerate value for the `external_id` field to schemas. Starting with the Account.

PR #137 follows with the rest of the schemas.

# Changes

- Add `omisego/ex_ulid` as a dependency of `EWalletDB`
- Create a new `EWalletDB.Types.ExternalID` to support this new Ecto custom type
- Add an `external_id/1` macro that allows easy adding of `ExternalID` to schemas.
- Add external_id to `Account` schema

# Implementation Details

## Background

As the number and complexity of schemas grow, we have lots of schemas that have their IDs in UUID format. Although UUIDs are easily recognisable and guarantees (high probability of) uniqueness, it does not describe what schema it identifies, and thus makes it hard to trace/determine what the ID represents.

## Format

```
"sym_" <> ulid

example> acc_01C9VBWZ92NC2ERZ24DFYS32PV
```

- `sym` = the schema's symbol, e.g. `mtk` for minted tokens. The symbol is always 3 characters

- `ulid` =  a generated ULID according to ULID [spec](https://github.com/ulid/spec)

This allows the external ID to be visually identifiable of its schema, as well as being sortable by creation time.

## Underscores as a separator

- Clearly separates the symbol from the content following it, making it easier to parse by the eyes.
- Does not rely on the string length. So if we need to add service/node address in the future, or any other data, we could use another underscore to separate the data e.g, `dsym_address_content` (`d` = decentralized?), `v1_sym_content` for versioning, etc.
- Text highlighting considers underscore a part of a word, not a word separator!

## Schemas with external IDs

- Account: `act_`
- API Key: `api_`
- Auth Token: `atk_`
- Access Key: `key_`
- Mint: `mnt_`
- Minted token: `tkn_`
- User: `usr_`
- Transaction request: `txr_`
- Transfer: `tfr_`

- Invite (disposable)
- Membership (internal)
- Role (uses `name`)
- Balance (uses `address`)

# Technical info

`ExULID.Ecto.ULID` is not used and instead `EWalletDB.Types.ExternalID` is created because the `ExULID` module generates pure ULIDs, while our use case is to have the External ID contains schema symbols along with the ULID, thus it had to be implemented on top of `ExULID.ULID` as Ecto string primitives.

The external_id is then applied to `Account` schema first. A separate PR will follow with all the rest.

Schema.get/2 still relies on internal ID because it needs to deal with permission checking and other code relies on it. It will also affect API requests. And so it could easily be its own PR.

# Usage

All responses that return accounts should return an `external_id` attribute for each returned object.

_To search for an account with a specific external ID, use `"search_terms": {"external_id": "acc_12345..."`_.

# Impact

Run `mix ecto.migrate` after code is deployed.


T-Dnzt:
@unnawut Aren't the external IDs supposed to replace the value of the IDs in all external communication? Having two IDs is quite confusing, how to know which one to use when I receive a JSON response?

unnawut:
@T-Dnzt Updated!

unnawut:
Also to be added to this PR: external IDs all in lower case

PR #137, opened 2018-03-30
External ID for APIKey, AuthToken, Key, Mint, TransactionRequest, TransactionConsumption, Transfer, User

unnawut:
Issue/Task Number: T147

# Overview

This PR continues from #134 where now all schemas that will have external IDs are implemented.

# Changes

- Create a migration that adds `external_id` fields to all affected schemas
- Add `external_id` field to those schemas
- Update all tests that need to return `external_id`

# Implementation Details

Same approach as #134 but covers the rest of the schemas.

## Schemas with external IDs

- Account: `acc_`
- API Key: `api_`
- Auth Token: `atk_`
- Access Key: `key_`
- Mint: `mnt_`
- Minted token: `tkn_`
- User: `usr_`
- Transaction request: `txr_`
- Transfer: `tfr_`

- Invite (disposable)
- Membership (internal)
- Role (uses `name`)
- Balance (uses `address`)

# Usage

All schemas that have external ID should have its object responses return `external_id` as well.

# Impact

Run `mix ecto.migrate` after the PR.


unnawut:
To be updated:
- MintedToken uses `tkn_ABC_01C9VBWZ92NC2ERZ24DFYS32PV ` where `tkn` is the schema minted_token and `ABC` is the minted token symbol

unnawut:
Closed in favor of #149

PR #143, opened 2018-04-04
Switch to use omisegoimages/ewallet-base to build Docker

sirn:
We now use custom base image at [omisegoimages](https://hub.docker.com/r/omisegoimages) to speed up builds and to allow for more stable base.


unnawut:
Did you mean to leave the reviewers blank?

sirn:
@unnawut nope :(

PR #144, opened 2018-04-05
T12 - Top level errors

T-Dnzt:
Issue/Task Number: T12

# Overview

This PR refactors the error logic to be shared between the two APIs.

# Changes

- Extract the shared logic (and refactor it) into a new ErrorHandler
- Keep the current error handlers in the APIs and delegate the calls to the shared ErrorHandler
- Give the possibility for shared errors as well as API specific ones.
- Works well with the current error listing

# Implementation Details

- The new error handler handle errors in a better and more unified way. There are four ways to use it (changeset, existing code, custom code and with template).
- A simple template system has been added to avoid adding new functions for each custom error description.

### DIFF
--- a/apps/ewallet/test/ewallet/web/v1/serializers/websocket_response_serializer_test.exs
+++ b/apps/ewallet/test/ewallet/web/v1/serializers/websocket_response_serializer_test.exs
@@ -1,0 +1,96 @@
+defmodule EWallet.Web.V1.WebsocketResponseSerializerTest do
+  use EWallet.Web.SerializerCase, :v1
+  alias EWallet.Web.V1.WebsocketResponseSerializer
+  alias Phoenix.Socket.Reply
+  alias Phoenix.Socket.Message
+
+  describe "serialize/3" do
+    test "serializes a websocket message" do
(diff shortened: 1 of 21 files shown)


apps/ewallet/test/ewallet/web/v1/serializers/websocket_response_serializer_test.exs
```diff
@@ -0,0 +1,96 @@
...
+      }
+    end
+  end
+
+  describe "encode!/1 with %Message{}" do
+    test "encodes fields" do
+      msg = %Message{ref: 1, topic: "topic", event: "event", payload: %{}}
+      {:socket_push, :text, encoded} = WebsocketResponseSerializer.encode!(msg)
+      decoded = Poison.decode!(encoded)
+      # refute true
```
unnawut: eh?

T-Dnzt: It's a test to see if you review dee dee :trollface:

unnawut: 🙄

PR #145, opened 2018-04-05
Add transaction consumption current count

T-Dnzt:
Issue/Task Number: T198

# Overview

@mederic-p requested a way to see how many consumptions had been made for a request. This PR fixes that.

# Changes

- Add `current_consumptions_count`

### DIFF
--- a/apps/ewallet/lib/ewallet/web/v1/serializers/transaction_request_serializer.ex
+++ b/apps/ewallet/lib/ewallet/web/v1/serializers/transaction_request_serializer.ex
@@ -9,7 +9,7 @@ defmodule EWallet.Web.V1.TransactionRequestSerializer do
   alias EWalletDB.TransactionRequest
 
   def serialize(%TransactionRequest{} = transaction_request) do
-    transaction_request = Repo.preload(transaction_request, :minted_token)
+    transaction_request = Repo.preload(transaction_request, [:minted_token, :consumptions])
 
     %{
       object: "transaction_request",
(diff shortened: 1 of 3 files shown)


apps/ewallet/lib/ewallet/web/v1/serializers/transaction_request_serializer.ex
```diff
@@ -25,6 +25,7 @@ defmodule EWallet.Web.V1.TransactionRequestSerializer do
       user_id: transaction_request.user_id,
       account_id: transaction_request.account_id,
       require_confirmation: transaction_request.require_confirmation,
+      current_consumptions_count: length(transaction_request.consumptions),
```
unnawut: This preloads all consumption records then count them though. Can we do a count query?

```ex
current_consumptions_count: Repo.one(from c in TransactionConsumption, select: count(c.id), where: c.transaction_request_id == ^transaction_request.id)
```

Feel free to move the query to `TransactionRequest.count_consumptions(transaction_request)` or something like that :)

T-Dnzt: @unnawut The `preload` will load everything in one request though (because we preload `minted_token` anyway) so I'm not sure it's worth it to do a separate count query. With that approach, we only have less stuff in memory. What do you think?

unnawut: Oh ok. Leave it as is then.

PR #149, opened 2018-04-09
Add id and uuid to schemas

unnawut:
Issue/Task Number: T147

# Overview

This PR continues from #134 where now all schemas that will have external IDs are implemented. This PR also replaces #137 with a new & cleaner approach.

**Review guide:** 
1. Review migration files first 
2. Review schema changes (mostly schema definitions and `get()` functions)
3. The rest should mostly be changing of field names (and preloading if needed).

# Changes

- Create a migration that renames `id` fields to `uuid` for all schemas
- Create a migration that renames `relation_id` to `relation_uuid` to be consistent
- Create a migration that adds new `id` fields with ExternalID type to all affected schemas
- Update all code that relies `id` to be `UUID` to use the new `ExternalID` type

# Implementation Details

Storing UUID as `id` and ExternalID as `external_id` has been complicate. This requires all code to convert `id` from external requests to `external_id`, including all relation_ids. It's requires mental note of which `id` to use and when it is or isn't transformed everywhere. This gets even more complicated when trying to save data back. The save also needs to know that the `id` maps back to `external_id`, etc. Multiply that by the number of schemas that needs to be aware of all this.

To have automatic mapping from `external_id` to `id` is difficult too. See https://github.com/elixir-ecto/ecto/issues/405. And in the end, `external_id` is used most anyway which actually works as `id` for all requests and most part of the code. 

So instead, this PR stores ExternalID, which is most frequently used  as `id` (sym_ULID) and the UUID as `uuid`. This way it's also clear that `account.parent_uuid` is always `parent.uuid`, and `account.parent_id` is always `parent.id`.

The `uuid` field should be used primarily for direct DB operations, e.g. primary keys, DB queries, joins, foreign keys, etc. While the `id` field should be used by controllers, gates, etc.

This way, most of the code should still be accessing ids via `id` and the value `sym_ULID` is being passed around most of the time. Even though a lot of file changes but most are simple renames and removing UUID format validation.

## Schemas with external IDs

- Account: `acc_`
- API Key: `api_`
- Auth Token: `atk_`
- Access Key: `key_`
- Mint: `mnt_`
- Minted token: `tkn_`
- User: `usr_`
- Transaction request: `txr_`
- Transfer: `tfr_`

- Invite (disposable)
- Membership (internal)
- Role (uses `name`)
- Balance (uses `address`)

# Usage

Technically this should not change any external requests/responses. Only that `id` fields should now be in External ID format.

# Impact

Run `mix ecto.migrate` after the PR.

The breaking changes are: 
1) If the external parties store the `id`s in their environment, they need to do mapping between the old `id` to `uuid` and prepare to receive `id` in ExternalID format.
2) Avatar file urls will break due to old id being part of the path. File renaming is not included.


unnawut:
`develop` keeps moving away 😢

apps/ewallet_db/lib/ewallet_db/helpers/assoc.ex
```diff
@@ -0,0 +1,29 @@
+defmodule EWalletDB.Helpers.Assoc do
```
T-Dnzt: 🎉 👍 🎉

apps/ewallet_db/lib/ewallet_db/transaction_request.ex
```diff
@@ -243,22 +241,22 @@ defmodule EWalletDB.TransactionRequest do
...
-                                                            {:error, Map.t}
+  @spec max_consumptions_reached?(%TransactionRequest{}, list(%TransactionConsumption{})) ::
+    true | false
   defp max_consumptions_reached?(request, consumptions) do
     limited_consumptions?(request) &&
     length(consumptions) >= request.max_consumptions
   end
 
-  @spec expire_if_max_consumption(%TransactionRequest{}) :: true | false
+  @spec limited_consumptions?(%TransactionRequest{}) :: true | false
```
T-Dnzt: 🙇

apps/ewallet_db/lib/ewallet_db/types/external_id.ex
```diff
@@ -17,6 +17,9 @@ defmodule EWalletDB.Types.ExternalID do
   alias Ecto.Schema
   alias ExULID.ULID
 
+  # 3-char symbols, 1-char underscore, 26-char ULID
```
T-Dnzt: 🙇

apps/ewallet_db/priv/repo/migrations/20180405191324_rename_id_to_uuid.exs
```diff
@@ -0,0 +1,33 @@
+defmodule EWalletDB.Repo.Migrations.RenameIdToUuid do
```
T-Dnzt: 👏

apps/ewallet_db/lib/ewallet_db/minted_token.ex
```diff
@@ -3,14 +3,18 @@ defmodule EWalletDB.MintedToken do
...
   alias Ecto.UUID
   alias EWalletDB.{Repo, Account, MintedToken}
 
-  @primary_key {:id, Ecto.UUID, autogenerate: true}
+  @primary_key {:uuid, UUID, autogenerate: true}
 
   schema "minted_token" do
+    external_id prefix: "tok_"
+
     field :friendly_id, :string # "EUR:123"
```
T-Dnzt: I think this should be called `id` now 🤔 💭

T-Dnzt: And merge friendly_id into external_id :)

unnawut:
@T-Dnzt Please review again

unnawut:
@T-Dnzt Please review again

PR #151, opened 2018-04-09
Make Admin API Keys optional

T-Dnzt:
Issue/Task Number: T179

# Overview

This PR makes the admin API Keys optional (disabled by default).

# Changes

- Add a new ENV variable: `ENABLE_ADMIN_CLIENT_AUTH`, `"false"` by default.
- It stays enabled for the tests, but some tests were added to test it.

### DIFF
--- a/apps/admin_api/config/config.exs
+++ b/apps/admin_api/config/config.exs
@@ -9,7 +9,8 @@ use Mix.Config
 config :admin_api,
   namespace: AdminAPI,
   ecto_repos: [],
-  sender_email: System.get_env("SENDER_EMAIL") || "admin@localhost"
+  sender_email: System.get_env("SENDER_EMAIL") || "admin@localhost",
+  enable_client_auth: System.get_env("ENABLE_ADMIN_CLIENT_AUTH") || "false"
 
 # Configs for the endpoint
 config :admin_api,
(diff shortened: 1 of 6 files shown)


apps/admin_api/config/config.exs
```diff
@@ -9,7 +9,8 @@ use Mix.Config
 config :admin_api,
   namespace: AdminAPI,
   ecto_repos: [],
-  sender_email: System.get_env("SENDER_EMAIL") || "admin@localhost"
+  sender_email: System.get_env("SENDER_EMAIL") || "admin@localhost",
+  enable_client_auth: System.get_env("ENABLE_ADMIN_CLIENT_AUTH") || "false"
```
unnawut: Can we convert "true" to boolean here so we don't need to worry about string vs. boolean downstream? E.g.

```ex
  enable_client_auth: System.get_env("ENABLE_ADMIN_CLIENT_AUTH") == "true" || false
```

apps/admin_api/lib/admin_api/v1/plugs/user_auth_plug.ex
```diff
@@ -18,29 +18,48 @@ defmodule AdminAPI.V1.UserAuthPlug do
 
   def init(opts), do: opts
 
-  def call(conn, _opts) do
+  def call(conn, opts) do
+    auth = Keyword.get(opts, :enable_client_auth,
```
unnawut: Can this go into `init/1` instead? Because it looks like option initialization to me. E.g.

```ex
  def init(opts) do
    Keyword.put_new(opts, :enable_client_auth, Application.get_env(:admin_api, :enable_client_auth))
  end

  def call(conn, opts) do
    if opts[:enable_client_auth] do
      conn
      |> parse_header()
      |> authenticate()
    else
      assign(conn, :authenticated, :client)
    end
  end
```

PR #152, opened 2018-04-10
Allow automatic Erlang node discovery

sirn:
# Overview

Allow automatic Erlang node discovery via [Peerage](https://github.com/mrluc/peerage).

# Changes

- Allow `NODE_HOST`, `ERLANG_COOKIE`, `NODE_DNS` and `NODE_NAME` to be configured.
- Add and configure Peerage.

# Implementation Details

Peerage and [libcluster](https://github.com/bitwalker/libcluster/) were being considered. Peerage is chosen for its configuration simplicity (12FA-compatible) and DNS discovery support which make it pretty much agnostic to any environment it will be deployed to.

Note that this is still very experimental.

# Usage

See README.


sirn:
Should be ready now since staging seems to be working.

README.md
```diff
@@ -146,13 +146,13 @@ Feel free to look around!
 We now need to pull the Elixir dependencies:
 
 ```
-mix deps.get
+$ mix deps.get
```
T-Dnzt: This `$` might make things clearer, but it's way more annoying to copy. I don't think it's worth it - where else will people copy command lines?

README.md
```diff
@@ -272,6 +272,93 @@ the credentials provided during the seed.
 
 All set! Start playing around with the API using the Swagger UI below to learn more about the available endpoints. Enjoy!
 
+### Clustering
```
T-Dnzt: Since this is an advanced topic, I believe it should be in a specific file in `/docs/` and referenced in the readme.

README.md
sirn: It's a matter of start copying from "mix" rather than double clicking (which is dangerous, because of double click will copy newline and it will make a command execute on paste), and some commands require manually changing values anyway.

IMHO preventing people from blinding copy & pasting should considered a good thing.

(Actually, have you noticed that all internal repo has `$` prefix in command examples?)

sirn:
@T-Dnzt fixed except the `$ ` stuff. Not sure if you still want it removed or not.

PR #155, opened 2018-04-19
Run Quantum in global mode

T-Dnzt:
Issue/Task Number: T216

# Overview

Quantum has an option to work within a cluster and avoid duplicated jobs. This PR enables it.

# Changes

- Enable `global` mode for Quantum


unnawut:
Adding links here just for future reference:
- https://hexdocs.pm/quantum/configuration.html#run-quantum-as-a-global-process
- https://hexdocs.pm/quantum/run-strategies.html

> When you have a cluster of nodes, you may not want same jobs to be generated on every single node, e.g. jobs involving db changes.
>
> In this case, you may choose to run Quantum as a global process, thus preventing same job being run multiple times because of it being generated on multiple nodes.

PR #160, opened 2018-04-23
Fix seeds failing due to id/uuid misreference

unnawut:
Issue/Task Number: T218

# Overview

Following #149, @jarindr found out that `mix seed` fails due to id/uuid references. This PR updates all seed scripts to correctly use `id` vs. `uuid` during the seed.

# Changes

- Update seed scripts to use `id` when displaying info, and use `uuid` and `relation_uuid` when doing direct inserts to the database.

# Implementation Details

To follow #149 correctly, the code interacting directly with the schema or the database need to use `uuid` and/or `relation_uuid`. While other code e.g. when displaying inserted data, the `id` should be used.

# Usage

Run `mix seed` and `mix seed --sample` should be successful.

# Impact

Deploy as usual.

### DIFF
--- a/apps/ewallet/lib/ewallet/web/preloader.ex
+++ b/apps/ewallet/lib/ewallet/web/preloader.ex
@@ -19,9 +19,9 @@ defmodule EWallet.Web.Preloader do
   @doc """
   Preloads associations into the given record(s).
   """
-  @spec preload(Ecto.Schema.t() | Ecto.Schema.t(), list(atom)) ::
+  @spec preload(Ecto.Schema.t() | Ecto.Schema.t(), atom() | [atom()]) ::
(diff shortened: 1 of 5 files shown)


unnawut:
@T-Dnzt Can't remember where I mentioned this (probably in one of the seed PRs) that I tried to add tests for the seed, but since they are scripts (`exs`), it becomes very difficult to control the environment for testing, e.g. cannot wrap the scripts inside transactions to run in parallel with other tests, difficult to suppress outputs so they don't jumble up test results, etc.

So I'm thinking this is better pushing up the priority for a seed refactor into proper modules then add tests there. :S

unnawut:
Turns out sirn is converting those seed scripts into proper modules...

PR #162, opened 2018-04-23
Generate and serve ex_doc via /docs

unnawut:
Issue/Task Number: T112

# Overview

Auto-generate docs during build process and serve at `/docs`.

# Changes

- Add `mix docs --output=public/docs` command to Dockerfile
- Add `/docs` to UrlDispatcher to serve static assets from `/public/docs`

# Implementation Details

N/A

# Usage

- On local machines, run `mix docs --output=public/docs` then browse to `/docs`
- On deployed servers, simply browse to `/docs`

image

# Impact

Deploy as usual


unnawut:
@T-Dnzt Not sure you'll like this solution (images are now in `assets/` in the root directory) but it's the simplest, producing the friendliest image urls, and works both on GitHub README and the generated docs.

i.e. `https://github.com/omisego/ewallet/blob/assets/provider_setup.jpg` and `http://10.5.10.10:4000/docs/assets/omgshop.png`

PR #178, opened 2018-05-10
Fix minted token's association with master account

### LINKED ISSUE #177: Minted token create
# Minted token create request error
Creating minted tokens response returns that the request is bad.  I have tried in swagger, admin panel and go. Even though I am using the default parameters it results in the following response:
```
{
  "version": "1",
  "success": false,
  "data": {
    "object": "error",
    "messages": null,
    "description": "Invalid parameter provided",
    "code": "client:invalid_parameter"
  }
}
```

# Your environment
I am on commit 169bf5c.

# Steps to reproduce
Parameters:
```
{
  "symbol": "BTC",
  "name": "Bitcoin",
  "description": "desc",
  "subunit_to_unit": 100
}
```

unnawut:
Issue/Task Number: T271

# Overview

Fixes `"client:invalid_parameter"` error when trying to create a minted token with `/admin/api/minted_token.create`. Fixes #177.

Depends on #179 otherwise the test with `enable_client_auth: false` is not possible.

# Changes

- Fix  `/minted_token.create`'s account association to the master account

# Implementation Details

This bug follows #151 that the client authentication (using API ID and API key) in AdminAPI can now be optional. This means `conn.assigns.account` assigned during `ClientAuthPlug` may not always be available, but `/minted_token.create` still relies on it.

Since we have no use case for minted token in non-master accounts, it's best to fix it to master account until we have a solid reason not to.

# Usage

Requesting `/minted_token.create` with valid parameters should now be successful.

# Impact

Deploy as usual.

### DIFF
--- a/apps/admin_api/lib/admin_api/v1/controllers/minted_token_controller.ex
+++ b/apps/admin_api/lib/admin_api/v1/controllers/minted_token_controller.ex
@@ -8,6 +8,7 @@ defmodule AdminAPI.V1.MintedTokenController do
   alias EWallet.Web.{SearchParser, SortParser, Paginator}
   alias EWalletDB.{Account, MintedToken}
   alias Ecto.UUID
+  alias Plug.Conn
 
   # The field names to be mapped into DB column names.
   # The keys and values must be strings as this is mapped early before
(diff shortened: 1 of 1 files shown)


unnawut:
I'm going to rethink this one a bit. I coupled the solutions for testing `:enable_client_auth` and the `/minted_token.create` bug which should be independent from each other.

unnawut:
The changes for this one remains the same actually. Just added typespecs in.

PR #179, opened 2018-05-10
Improvements to ClientAuthPlug and UserAuthPlug

unnawut:
Issue/Task Number: T271

# Overview

Small refactors and improvements of `ClientAuthPlug` and `UserAuthPlug` and its tests to cover all `:enable_client_auth` cases.

# Changes

- ~Move `:enable_client_auth` check from `init/1` to `call/2` to support runtime config change~ No need to. I was coupling #178 and #179 together when I was doing that.
- Add tests to `ClientAuthPlugTest` and `UserAuthPlugTest` to cover all `:enable_client_auth` combinations
- Update moduledoc for `ClientAuthPlug` & `UserAuthPlug`
- Refined `UserAuthPlug.call/2` to be more specific when handling different kinds of auth headers:

# Implementation Details

- Added more `ClientAuthPlugTest` and `UserAuthPlugTest` tests. This should cover all combinations of `enable_client_auth: true` v.s. `false` and 4-part v.s. 2-part Authorization header.
- The refined `UserAuthPlug.call/2` can handle different kinds of auth headers more accurately:
  - When `enable_client_auth: true`
    - Attempts to authenticate `api_key_id:api_key:user_id:auth_token` header
    - Other header formats will not be authenticated
  - When `enable_client_auth: false`
    - Attempts to authenticate `user_id:auth_token` header
    - If `api_key_id:api_key:user_id:auth_token` header is provided, it authenticates both `api_key_id:api_key` and `user_id:auth_token`. Providing incorrect `api_key_id:api_key` will result in `client:invalid_api_key` error.
- ☝️ This refined plug also allows for more accurate error codes. Before this PR an invalid `user_id:auth_token` would return `client:invalid_api_key` error. With this PR it would return `client:invalid_api_key` and `user:access_token_not_found` accurately.

# Usage

Run `mix test`

# Impact

Deploy as usual.

### DIFF
--- a/apps/admin_api/config/test.exs
+++ b/apps/admin_api/config/test.exs
@@ -1,12 +1,13 @@
 use Mix.Config
 
+# General application configuration
+config :admin_api, enable_client_auth: true
+
 # We don't run a server during test. If one is required,
 # you can enable the server option below.
 config :admin_api, AdminAPI.Endpoint,
   secret_key_base: "G1DLBdjjJSoSiQRa5Gf8YrWUx5yrX+JFmZx+UBk829W1+e0oJ9TYrW/GkIgrAdfm",
   server: false
-
-config :admin_api, enable_client_auth: "true"
 
 # Configs for Bamboo emailing library
 config :admin_api, AdminAPI.Mailer, adapter: Bamboo.TestAdapter
--- a/apps/admin_api/lib/admin_api/v1/plugs/client_auth_plug.ex
+++ b/apps/admin_api/lib/admin_api/v1/plugs/client_auth_plug.ex
@@ -2,8 +2,24 @@ defmodule AdminAPI.V1.ClientAuthPlug do
   @moduledoc """
   This plug checks if valid `api_key_id` and `api_key` are provided.
 
-  If `api_key_id` and `api_key` matches the database record, the plug assigns
-  the `api_key_id` and `account` to the connection along with `authenticated: true`.
+  On success, the plug assigns the following `conn.assigns`:
+
+    - `authenticated`: Set to `:client` to indicate that the request has been authenticated
+                       at the client level.
+    - `api_key_id`: The API key used to authenticate the request.
+    - `auth_account`: The account that is associated with the API key.
+
+  If client auth is disabled, the plug assigns the following `conn.assigns`:
+
+    - `authenticated`: Set to `:client`.
+    - `api_key_id`: Not assigned.
+    - `auth_account`: Not assigned.
+
+  On failure, the plug halts, then assigns the following `conn.assigns`:
+
+    - `authenticated`: Set to `false`.
+    - `api_key_id`: Not assigned.
+    - `auth_account`: Not assigned.
   """
   import Plug.Conn
   import AdminAPI.V1.ErrorHandler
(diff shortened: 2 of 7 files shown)


apps/admin_api/lib/admin_api/v1/controllers/minted_token_controller.ex
```diff
@@ -53,7 +53,7 @@ defmodule AdminAPI.V1.MintedTokenController do
   @doc """
   Creates a new Minted Token.
   """
-  def create(%{assigns: %{account: account}} = conn, attrs) do
+  def create(%{assigns: %{auth_account: account}} = conn, attrs) do
```
T-Dnzt: Aren't we removing this entirely in PR #178?

apps/admin_api/config/test.exs
```diff
@@ -1,12 +1,13 @@
 use Mix.Config
 
+# General application configuration
+config :admin_api, enable_client_auth: true
```
T-Dnzt: How can you define this in an ENV variable if it expects a boolean? Having a check `== "true"` would break @sirn's request about not using logic for ENV variables, no?

apps/admin_api/lib/admin_api/v1/plugs/user_auth_plug.ex
```diff
@@ -1,32 +1,32 @@
...
+    - `user`: Not assigned.
   """
   import Plug.Conn
   import AdminAPI.V1.ErrorHandler
   alias EWalletDB.Helpers.Crypto
   alias EWalletDB.{AuthToken, User}
   alias AdminAPI.V1.ClientAuthPlug
 
-  def init(opts) do
-    Keyword.put_new(
```
T-Dnzt: I think I remember someone telling me to move it from `call` to `init` :trollface:

apps/admin_api/lib/admin_api/v1/plugs/client_auth_plug.ex
```diff
@@ -12,20 +28,21 @@ defmodule AdminAPI.V1.ClientAuthPlug do
...
-      "true" ->
+    case enable_client_auth?(opts) do
+      true ->
         conn
         |> parse_header()
         |> authenticate()
 
       _ ->
-        assign(conn, :authenticated, :client)
+        assign(conn, :authenticated, false)
```
T-Dnzt: Are we sure this doesn't break anything anywhere? Worried we're checking the `authenticated` somewhere and expect it to be true :o

apps/admin_api/lib/admin_api/v1/controllers/minted_token_controller.ex
unnawut: Yes but they're separate PR so I can't remove this patten matching here.

apps/admin_api/lib/admin_api/v1/plugs/user_auth_plug.ex
unnawut: Yeah sorry. It was a for a different reason though. I think this is needed so it can be tested with different configs 🙏

unnawut:
I'm going to rethink this one a bit. I coupled the solutions for testing `:enable_client_auth` and the `/minted_token.create` bug which should be independent from each other.

apps/admin_api/config/test.exs
sirn: Should be fine, DeferredConfig can normalize settings.

unnawut:
@T-Dnzt @sirn This is ready for another review

PR #185, opened 2018-05-11
Add max_consumption_per_user for transaction requests

T-Dnzt:
Issue/Task Number: T229

# Overview

This PR adds a way to specify that a transaction request should be consumed a specific amount of time per user (usually 1 in the case of a giveaway).

# Changes

- Add `max_consumptions_per_user` to transaction requests
- Ensure that value has not been reached by the current user (set on the balance)
- Add tests 
- Update Swagger spec

### DIFF
--- a/apps/ewallet/test/ewallet/gates/transaction_consumption_gate_test.exs
+++ b/apps/ewallet/test/ewallet/gates/transaction_consumption_gate_test.exs
@@ -452,6 +452,145 @@ defmodule EWallet.TransactionConsumptionGateTest do
       assert error == :expired_transaction_request
     end
 
+    test "returns a 'max_consumptions_per_user_reached' error if the maximum number of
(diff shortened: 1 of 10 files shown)


apps/ewallet/test/ewallet/gates/transaction_consumption_gate_test.exs
```diff
@@ -452,6 +452,145 @@ defmodule EWallet.TransactionConsumptionGateTest do
...
+          "metadata" => nil,
+          "idempotency_token" => "1234",
+          "token_id" => nil
+        })
+
+      assert res == :error
+      assert error == :max_consumptions_per_user_reached
+    end
+
+    test "allows only one consume per user with four consumes at the same time", meta do
```
unnawut: :mindblown: 🎖🥇🥈🍾 🌟

PR #190, opened 2018-05-15
Cleanup Consumption and handle idempotency

T-Dnzt:
Issue/Task Number: T293

# Overview

UPDATE: This has turn out into a full refactor of the consumption gate which was way too big. 

It is now split into two different gates (`consumer` and `confirmer`) and two helpers (a fetcher and a validator). A fetcher was also extracted from the `TransactionRequestGate`.

# Changes

- Validate `max_consumptions_per_user` correctly
- Ensure tests are testing the flow correctly
- Set the correct tasks to be run with Quantum
- Extract consumption validations into its own module
- Extract consumption confirmation step into its own gate
- Fix tests and reorganize files


apps/ewallet/lib/ewallet/gates/transaction_consumption_gate.ex
```diff
@@ -144,21 +144,55 @@ defmodule EWallet.TransactionConsumptionGate do
...
-        {:ok, balance}
-
-      max ->
-        case TransactionConsumption.all_active_for_user(balance.user_uuid) do
-          [] -> {:ok, balance}
-          _ -> {:error, :max_consumptions_per_user_reached}
-        end
+    validate_max_consumptions_per_user(
+      max: request.max_consumptions_per_user,
+      is_user_request: !is_nil(balance.user_uuid) && is_nil(balance.account_uuid),
```
unnawut: Can we convert this to a new Balance function: `Balance.is_user_balance?(balance)`?

I think this might be used somewhere else in the future and the logic to check this might change.

apps/ewallet/test/ewallet/gates/transaction_consumption_gate_test.exs
```diff
@@ -191,6 +191,49 @@ defmodule EWallet.TransactionConsumptionGateTest do
 
       assert res == {:error, :account_balance_mismatch}
     end
+
+    test "works for account even if max_consumptions_per_user is set", meta do
```
unnawut: What's the rationale for accounts bypassing this rule? Just for my own knowledge.

T-Dnzt:
The consumption code is too messy, I'm going to refactor it in this PR.

apps/ewallet/test/ewallet/gates/transaction_consumption_gate_test.exs
T-Dnzt: Good point, removed it in the refactor.

apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
```diff
@@ -0,0 +1,132 @@
...
+  end
+
+  @spec validate_before_confirmation(TransactionConsumption.t(), Account.t() | User.t()) ::
+          {:ok, TransactionConsumption.t()}
+          | {:error, Atom.t()}
+  def validate_before_confirmation(consumption, owner) do
+    with {request, balance} <- {consumption.transaction_request, consumption.balance},
+         true <-
+           TransactionRequest.is_owned_by?(request, owner) || :not_transaction_request_owner,
+         true <-
```
mederic-p: 👎 (duplicated)

apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
```diff
@@ -0,0 +1,132 @@
...
+
+      false ->
+        case amount do
+          nil -> {:ok, request, request.amount}
+          _amount -> {:error, :unauthorized_amount_override}
+        end
+    end
+  end
+
+  # defp get_and_validate_minted_token(request, nil) do
```
mederic-p: Do we need to let this here?

apps/ewallet/test/ewallet/fetchers/transaction_consumption_fetcher_test.exs
```diff
@@ -0,0 +1,78 @@
...
+      minted_token: minted_token,
+      receiver_balance: receiver_balance,
+      sender_balance: sender_balance,
+      account_balance: account_balance,
+      request: transaction_request
+    }
+  end
+
+  describe "get/1" do
+    test "returns the consumption do when given valid ID", meta do
```
mederic-p: remove `do`

apps/ewallet/test/ewallet/validators/transaction_consumption_validator_test.exs
```diff
@@ -0,0 +1,3 @@
+defmodule EWallet.TransactionConsumptionValidatorTest do
```
unnawut: :imwatchingyou:

apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
```diff
@@ -35,14 +35,12 @@ defmodule EWallet.TransactionConsumptionValidator do
...
-         true <-
-           TransactionRequest.is_owned_by?(request, owner) || :not_transaction_request_owner,
          true <-
            TransactionRequest.is_owned_by?(request, owner) || :not_transaction_request_owner,
          {:ok, request} <- TransactionRequest.expire_if_past_expiration_date(request),
+         {:ok, _balance} <- validate_max_consumptions_per_user(request, balance),
          true <- TransactionRequest.valid?(request) || request.expiration_reason,
-         {:ok, consumption} = TransactionConsumption.expire_if_past_expiration_date(consumption),
-         {:ok, _balance} <- validate_max_consumptions_per_user(request, balance) do
+         {:ok, consumption} = TransactionConsumption.expire_if_past_expiration_date(consumption) do
```
unnawut: Curious why the shuffle?

T-Dnzt: Hmmm, because the `max_consumptions_per_user` is a transaction request check, it made more sense to have it before reaching the validation for the consumption (especially since the following line checks if the consumption got expired or not). The `TransactionConsumption.expire_if_past_expiration_date` will actually never return an error so it should be the last line of the `with` :)

PR #204, opened 2018-05-23
Fix already started Endpoint error

T-Dnzt:
Issue/Task Number: T331

# Overview

I tried stopping the processes after the tests, but it doesn't seem to work (keeps saying the process isn't running anymore already).

So I just remove the `:ok` check.

# Changes

- Remove ` {:ok, _} = ` when starting `TestEndpoint`


unnawut:
​Maybe get @sirn to have a look at this PR as well?

T-Dnzt:
It's not fixed :'( Will take another look...

T-Dnzt:
This is not the way to fix it... closing this.

PR #205, opened 2018-05-23
/transaction.create endpoint in Admin API

T-Dnzt:
Issue/Task Number: T272

# Overview

Add `/transaction.create` endpoint in Admin API.

# Changes

- Add new endpoint to create transactions
- Update OpenAPI spec

### DIFF
--- a/apps/admin_api/lib/admin_api/v1/router.ex
+++ b/apps/admin_api/lib/admin_api/v1/router.ex
@@ -1,5 +1,6 @@
 defmodule AdminAPI.V1.Router do
   use AdminAPI, :router
+  alias EWallet.Web.V1.Plug.Idempotency
   alias AdminAPI.V1.{ClientAuthPlug, UserAuthPlug}
 
   # Pipeline for plugs to apply for all endpoints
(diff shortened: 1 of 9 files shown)


apps/admin_api/lib/admin_api/v1/plugs/idempotency.ex
```diff
@@ -0,0 +1,31 @@
+defmodule AdminAPI.V1.Plug.Idempotency do
```
unnawut: Can we merge with `EWalletAPI.V1.Plug.Idempotency` and move to `EWallet.Web.V1.IdempotencyPlug`?

apps/admin_api/lib/admin_api/v1/router.ex
```diff
@@ -33,6 +38,12 @@ defmodule AdminAPI.V1.Router do
     post("/transaction.all", TransactionController, :all)
     post("/transaction.get", TransactionController, :get)
 
+    scope "/" do
+      pipe_through([:idempotency])
+
+      post("/transaction.create", TransactionController, :create)
```
unnawut: This is the same logic as eWallet API's `/transfer` right? Thinking if we should have consistent naming...

T-Dnzt: Let's rename it in the eWallet API then.

apps/admin_api/lib/admin_api/v1/plugs/idempotency.ex
T-Dnzt: Tried it... But had some issues with it so I moved back to duplicated code.

PR #207, opened 2018-05-23
Use dedicated instance for Jenkins slave

sirn:
# Overview

Switch Jenkins to use dedicated auto-scaling preemptible node pool to make the build more predictable (resource-wise) and to reduce overall cost of running the CI. 

# Changes

- Add node tolerations and node selector in Jenkinsfile (to schedule to dedicated node).
- Add resource requests and limits.

# Implementation Details

N/A

# Usage

N/A

# Impact

N/A


sirn:
Caveat is it might take around 5 minutes for the build to start if there's no nodes in the preemptible pool (in which autoscaler will automatically spin-off a new one). This will be the case for the first build after some period where OmiseGO projects configured in Jenkins are inactive (e.g. no push).

unnawut:
YES THANK YOU!

One small question. This does not impact the time it takes for a build to show up in Jenkins right?

That is, when a branch has been pushed or PR submitted, it should show up in Jenkins near-immediately with a new build number. But just that the wait time until the build actually starts might be longer?

sirn:
Yup

PR #209, opened 2018-05-24
Add warnings_as_errors flag to Mixfile

unnawut:
Issue/Task Number: T312

# Overview

This PR adds warnings_as_errors flag as suggested by @InoMurko so we detect warnings in our code.

# Changes

- Add `elixirc_options: [warnings_as_errors: true]` to all app's Mixfiles

# Implementation Details

InoMurko originally suggested using `--warnings_as_errors` flag during compilation, but since we can add this as default in the Mixfile, I decided to add it to Mixfile instead so we can continue using `mix compile`

# Usage

Run `mix compile` with warnings such as unused variable or alias should raise an error.

# Impact

Deploy as usual

### DIFF
--- a/apps/admin_api/mix.exs
+++ b/apps/admin_api/mix.exs
@@ -11,6 +11,7 @@ defmodule AdminAPI.Mixfile do
       lockfile: "../../mix.lock",
       elixir: "~> 1.4",
       elixirc_paths: elixirc_paths(Mix.env),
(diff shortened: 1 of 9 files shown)


unnawut:
I'll wait for the pending PRs to merge before merging this one.

unnawut:
To merge after #217 & #219 are merged

PR #213, opened 2018-05-24
Fix double quotes in OpenAPI Spec

T-Dnzt:
Issue/Task Number: A

# Overview

Oops... Put some double quotes in the spec. somehow. Fixing it in this PR.


jarindr:
double, double quote 🥇

PR #214, opened 2018-05-25
Prevent random failure for consumption consumer gate

T-Dnzt:
Issue/Task Number: A

# Overview

Saw this test randomly failing yesterday:

```
  1) test consume/2 with user allows only one consume per user with four consumes at the same time (EWallet.TransactionConsumptionConsumerGateTest)
     test/ewallet/gates/transaction_consumption_consumer_gate_test.exs:652
     Assertion with == failed
     code:  assert response == :max_consumptions_per_user_reached
     left:  %{
              __meta__: #Ecto.Schema.Metadata<:loaded, "transaction_consumption">,
              __struct__: EWalletDB.TransactionConsumption,
```
This PR fixes that by checking that one of the responses is a valid consumption and the others are errors.


unnawut:
Ah that's because we can't guarantee which one gets executed first... ✅

PR #218, opened 2018-05-25
Remove searching capabilities from Wallet Controller

T-Dnzt:
Issue/Task Number: T346

# Overview

Currently, the way the searcher works isn't great and it was causing the `account.get_wallets` and `user.get_wallets` endpoints to return unwanted results. For now, this PR removes the search step.

### DIFF
--- a/apps/admin_api/test/admin_api/v1/controllers/wallet_controller_test.exs
+++ b/apps/admin_api/test/admin_api/v1/controllers/wallet_controller_test.exs
@@ -55,11 +55,14 @@ defmodule AdminAPI.V1.WalletControllerTest do
 
       wallets = response["data"]["data"]
       assert length(wallets) == 2
-      assert Enum.at(wallets, 0)["account_id"] == account.id
-      assert Enum.at(wallets, 0)["identifier"] == "burn"
 
-      assert Enum.at(wallets, 1)["account_id"] == account.id
-      assert Enum.at(wallets, 1)["identifier"] == "primary"
+      wallets =
+        Enum.map(wallets, fn wallet ->
+          {wallet["account_id"], wallet["identifier"]}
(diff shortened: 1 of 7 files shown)


apps/ewallet_db/lib/ewallet_db/wallet.ex
```diff
@@ -111,8 +111,12 @@ defmodule EWalletDB.Wallet do
     |> put_change(:encryption_version, Cloak.version())
   end
 
-  def all_for(%Account{} = account), do: from(t in Wallet, where: t.account_uuid == ^account.uuid)
-  def all_for(%User{} = user), do: from(t in Wallet, where: t.user_uuid == ^user.uuid)
+  def all_for(%Account{} = account),
```
unnawut: Sirn despises `..., do:` format that extends multiple line. Can we change to the usual multiliner?

T-Dnzt: It's the `mix format`, urgh...

apps/admin_api/test/admin_api/v1/controllers/wallet_controller_test.exs
```diff
@@ -69,8 +72,8 @@ defmodule AdminAPI.V1.WalletControllerTest do
       assert is_boolean(pagination["is_first_page"])
     end
 
-    test "returns a list of wallets according to search_term, sort_by and sort_direction" do
-      {:ok, account} = :account |> params_for() |> Account.insert()
+    test "returns a list of wallets according to sort_by and sort_direction" do
+      account = insert(:account)
```
unnawut: Note to self that `insert(:account)` here so the number of wallets are more manageable during these tests.

PR #219, opened 2018-05-29
Surround Transfer retrieval after insert with transaction

T-Dnzt:
Issue/Task Number: T347

# Overview

Since we have multiple nodes of the DB, there seem to be some randomly failing method in the Transfer insert flow. Basically, the retrieval after the insert fails because the record was "potentially" inserted in another node. This PR surrounds the insert/get flow with a transaction as an attempt to fix that.

# Changes

- Add `Repo.transaction()` around in `insert`

### DIFF
--- a/apps/ewallet_db/lib/ewallet_db/transfer.ex
+++ b/apps/ewallet_db/lib/ewallet_db/transfer.ex
@@ -172,16 +172,24 @@ defmodule EWalletDB.Transfer do
   using the passed idempotency token.
   """
   def insert(attrs) do
-    changeset = changeset(%Transfer{}, attrs)
     opts = [on_conflict: :nothing, conflict_target: :idempotency_token]
 
-    case Repo.insert(changeset, opts) do
-      {:ok, transfer} ->
-        {:ok, get_by_idempotency_token(transfer.idempotency_token)}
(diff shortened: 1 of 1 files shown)


apps/ewallet_db/lib/ewallet_db/transfer.ex
```diff
@@ -172,18 +172,28 @@ defmodule EWalletDB.Transfer do
...
-      {:ok, transfer} ->
-        {:ok, get_by_idempotency_token(transfer.idempotency_token)}
-
-      changeset ->
-        changeset
+    %Transfer{}
+    |> changeset(attrs)
+    |> do_insert(opts)
+    |> case do
+      {_, res} -> res
```
unnawut: Would the new return types cause problem? E.g. before, it would return `{:ok, result}` or `changeset`. Now it's `{:ok, result}` or `{:error, changeset}`?

T-Dnzt: `insert` actually returns `{:error, changeset}`, here it's just shortcuted to `changeset`. [Insert doc](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2)

PR #222, opened 2018-06-05
Fix idempotent retrieval attempt 

T-Dnzt:
Issue/Task Number: T346

# Overview

Retrieval of transactions after insert when a conflict occurred is still not working properly and failing once in a while. I'm getting desperate and adding a loop that will try reloading the record 3 times before sending back an error. Will try this in staging and see if we can improve it somehow.


apps/ewallet_db/lib/ewallet_db/transfer.ex
```diff
@@ -186,14 +186,34 @@ defmodule EWalletDB.Transfer do
...
+  defp attempt_retrieval(_transfer, _idempotency_token, 3) do
+    {:error, :inserted_transaction_could_not_be_loaded}
+  end
+
+  # Due to DB sync issues accross clusters and because of the way we retrieve
+  # the transactions (ignoring failure on conflicts), it's possible we're not able
+  # to retrive the record right away. We'll try 3 times in total, with a ms delay
+  # starting at 0, then 5, then 10ms before returning an error.
+  defp attempt_retrieval(nil, idempotency_token, count) do
+    Process.sleep(count * 5)
```
unnawut: Can we use a module annotation e.g. `@attempt_retrieval_sleep_ms 5`

unnawut:
Probably good to have sirn review this as well. Maybe he has other solutions.

sirn:
Can you provide steps to reproduce this error? I don't really think retrying is the right solution…

PR #226, opened 2018-06-06
321-forget-passworld

jarindr:
Issue/Task Number: 321-forget-passworld
# Overview
wireup logic for forget password 

# Changes

- refine some bugs
- wireup forget password logic

# Implementation Details

simple validate error most of the case on client for better UX implementation

# Usage

go to `/login` and play around with forget password

# Impact

deploy like usual

### DIFF
--- a/apps/admin_panel/assets/src/omg-create-account-modal/index.js
+++ b/apps/admin_panel/assets/src/omg-create-account-modal/index.js
@@ -65,7 +65,7 @@ class CreateAccountModal extends Component {
       this.setState({ stage: 'finished' })
     } else {
       this.setState({
-        error: result.data.data.code,
+        error: result.data.data.description,
         submitting: false
       })
     }
--- a/apps/admin_panel/assets/src/omg-page-api/index.js
+++ b/apps/admin_panel/assets/src/omg-page-api/index.js
@@ -46,6 +46,14 @@ const ApiKeyContainer = styled.div`
 `
 const KeySection = styled.div`
   max-width: 700px;
+  h3 {
+    font-size: 18px;
+    font-weight: 400;
+    margin-bottom: 20px;
+  }
+  p {
+    color: ${props => props.theme.colors.B100};
+  }
 `
 const KeySectionEwallet = KeySection.extend`
   margin-top: 50px;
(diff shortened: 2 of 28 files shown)


apps/admin_panel/assets/src/omg-create-account-modal/index.js
```diff
@@ -65,7 +65,7 @@ class CreateAccountModal extends Component {
       this.setState({ stage: 'finished' })
     } else {
       this.setState({
-        error: result.data.data.code,
+        error: result.data.data.description,
```
T-Dnzt: I think it would make sense to display both the error code and the description. Error codes are always `something_happened`, so you could beautify it into:

`Something happened`
`This is the description I got from the server`.

Or, another approach is to show the description first, like how you would see in some OS:

`This is the description I got from the server`
`Error code: "something_happened"` 

So users can just give us the error code :)

T-Dnzt: This can be changed everywhere in another PR btw!

apps/admin_panel/assets/src/omg-page-api/index.js
```diff
@@ -88,31 +96,37 @@ class ApiKeyPage extends Component {
...
       <KeySection>
         <h3>Admin API Key</h3>
-        {/* <p>
-          When asked, “If you could wish for one thing only, what would that wish be?” almost
-          everyone; from beauty pagent contestants, to politicians, to religious leaders, to
-          children,
-        </p> */}
+        <p>
+        The Admin API key is used to authenticate an API and allows that specific API to access various admin-related functions such as creating new minted tokens, mint more tokens, create and manage accounts, create new API keys, etc.
+        </p>
```
T-Dnzt: What are Admin API Keys exactly? A subset of client API keys? If those are disabled, I think you should be able to create them 🤔 

As a reminder, here are the types we have:
- Client API Keys (Either to access the Admin API before logging in, or to authenticate iOS/Android non-sensitive clients)
- Access/Secret Keys (Super powerful keys used in server apps to calls the eWallet API and do pretty much anything, should never be embedded in a client!)

apps/admin_panel/assets/src/omg-create-account-modal/index.js
jarindr: ideally, `description` should not be use in anycase since it probably need to be localized, but yeah showing error code is good too will change that

T-Dnzt: I agree with that, but we're not localizing yet. If you wanna go that way, duplicate the errors descriptions in the app the way the mobile apps did it and use the error code to find the right local description.

Issue #224, opened 2018-06-05
Randomly Failing to retrieve a transfer after insert


T-Dnzt:
# Subject of the issue

Randomly Failing to retrieve a transfer after insert in:

```elixir
  def insert(attrs) do
    changeset = changeset(%Transfer{}, attrs)
    opts = [on_conflict: :nothing, conflict_target: :idempotency_token]

    case Repo.insert(changeset, opts) do
      {:ok, transfer} ->
        {:ok, get_by_idempotency_token(transfer.idempotency_token)}

      changeset ->
        changeset
    end
  end
```

# Your environment

Staging environment, using Ruby SDK or iOS SDK.

# Steps to reproduce

1. Get Ruby SDK
2. Replace the config in `env.rb` with real values
3. Run 

```
~/src/ruby-sdk on  T305-renaming! ⌚ 16:05:55
$ rspec ./spec/omisego/wallet_spec.rb                                                                                                                                                                                            ‹ruby-2.3.3›

OmiseGO::Wallet
  .all
    retrieves the list of wallets
  .credit
    with valid params
      credits the user's wallet
    with params account_id and account_address
      credits the user's wallet (FAILED - 1)
  .debit
    with valid params
      debits the user's wallet (FAILED - 2)
    with params account_id and account_address
      debit/s the user's wallet

Failures:

  1) OmiseGO::Wallet.credit with params account_id and account_address credits the user's wallet
     Failure/Error: expect(wallets).to be_kind_of OmiseGO::List
       expected #<OmiseGO::Error:70175132363600 code: server:internal_server_error, description: Something went wrong on the server, messages: > to be a kind of OmiseGO::List
     # ./spec/omisego/wallet_spec.rb:67:in `block (5 levels) in <module:OmiseGO>'
     # /Users/thibault/.rvm/gems/ruby-2.3.3/gems/vcr-3.0.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
     # /Users/thibault/.rvm/gems/ruby-2.3.3/gems/vcr-3.0.3/lib/vcr.rb:189:in `use_cassette'
     # ./spec/omisego/wallet_spec.rb:55:in `block (4 levels) in <module:OmiseGO>'

  2) OmiseGO::Wallet.debit with valid params debits the user's wallet
     Failure/Error: expect(wallets).to be_kind_of OmiseGO::List
[...]

PR #227, opened 2018-06-06
Fixing Transfer retrieval after safe insert

T-Dnzt:
Issue/Task Number: T346

# Overview

Attempting to fix [this issue](https://github.com/omisego/ewallet/issues/224) using Repo.Multi. Removed the retries as it wasn't a good solution. The code can be improved but I want to test it in staging, see if it actually fixes anything or not. My next step will be to try reproducing in local with a cluster.

# Changes

- Use `Repo.Multi` to insert a Transfer and retrieve the latest inserted
- Ensure the local ledger can handle existing idempotency token by returning the existing entry instead or returning an error.

### DIFF
--- a/apps/ewallet/lib/ewallet/formatters/transfer_formatter.ex
+++ b/apps/ewallet/lib/ewallet/formatters/transfer_formatter.ex
@@ -4,7 +4,7 @@ defmodule EWallet.TransferFormatter do
   """
   def format(transfer) do
     %{
-      "correlation_id" => transfer.idempotency_token,
+      "idempotency_token" => transfer.idempotency_token,
       "metadata" => transfer.metadata,
       "token" => %{
         "id" => transfer.token.id,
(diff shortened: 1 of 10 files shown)


apps/ewallet/lib/ewallet/formatters/transfer_formatter.ex
```diff
@@ -4,7 +4,7 @@ defmodule EWallet.TransferFormatter do
   """
   def format(transfer) do
     %{
-      "correlation_id" => transfer.idempotency_token,
+      "idempotency_token" => transfer.idempotency_token,
```
mederic-p: 💯

T-Dnzt:
@unnawut It's something I can try, but the transaction should ensure it happens in the same process 🤔 We weren't using transaction before but I haven't tried calling outside the transaction. Let me try this out first and if things still don't work, I'll try your solution :)

PR #228, opened 2018-06-07
json specification 

InoMurko:
Issue/Task Number:
157 https://github.com/omisego/ewallet/issues/157
# Overview

ewallet provides a yaml api specification the requirements are that we also provide json file format specification.

# Changes

- JSON file serving

# Implementation Details

Following yaml implementation. 
Function callback in macro and catching exceptions.

# Usage

`java -jar bin/openapi-generator-cli.jar generate -i ewallet//apps/ewallet_api/priv/spec.yaml -g openapi -o output/`

# Impact

JSON specs can be added as a CI step (but since you're defining the yaml manually I don't think it's needed)


unnawut:
Note that I'm trying this PR out today. Will give feedbacks once done.

InoMurko:
Ok. Rebased for your convenience

unnawut:
Looking good! Thanks for keep finding the solution 🙇 

Some notes:

1. I got two errors when putting the generated JSON specs into https://editor.swagger.io:

```
Semantic error at components.schemas.UnpaginatedListSchema.properties.data
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 1100

Semantic error at components.schemas.PaginatedListSchema.properties.data
Schemas with 'type: array', require a sibling 'items: ' field
Jump to line 1116
```

2. One small test failure at the moment:

```ex
  1) test /errors endpoints returns the json spec when calling /errors.json (EWallet.Web.APIDocs.ControllerTest)
     test/ewallet/web/api_docs/controller_test.exs:82
     Assertion with =~ failed
     code:  assert conn.resp_body() =~ ~r"\\\"- token_not_found:\\\\n    code:"
     left:  "\"- invalid_access_secret_key:\\n    code: ..."
     right: ~r/\"- token_not_found:\\n    code:/
     stacktrace:
       test/ewallet/web/api_docs/controller_test.exs:88: (test)
```

3. @sirn Any opinions on having a new build dependency: `openapi-generator-cli.jar`? Since this is only needed during builds, we could install from Github release -> generate JSON -> remove installed binaries, all within the [`Dockerfile`](https://github.com/omisego/ewallet/blob/develop/Dockerfile)?

unnawut:
Ignore my feedback about the JSON spec errors. Looks like the YAML specs cause those errors as well (and even more). I'll deal with that as a separate PR.

apps/ewallet/lib/ewallet/web/api_docs/errors.json.eex
```diff
@@ -0,0 +1,10 @@
+<%= for {key, error} <- assigns.errors do %>
```
unnawut: Looks like this is still generating the yaml format?

unnawut:
#237 is merged. The online editor errors should be fixed now.

Please try rebase with the latest change from `develop` branch. :)

unnawut:
Looks good to go for me from implementation point of view.

@sirn Any comments on directly including the jar executable in the repo?

InoMurko:
just to clarify why I included the jar in the repo. The idea was that we could test  yaml and json specs by generating a client lib. The client lib would then be used in acceptance tests. I’m positive robotframework allows us to do that.
But since we're not there yet, we can also decide to remove it.

T-Dnzt:
Having the JSON file directly in the repo bothers me a bit, I was hoping we could just generate it on the fly and return it from the controller 🤔 We would need to re-generate it every time a change is made to one of the `spec.yaml` which might get forgotten along the way.

InoMurko:
Invoking the jar on every request? 
The other option is being a part of deployment procedure (https://hexdocs.pm/distillery/boot-hooks.html pre start hook?)

unnawut:
I like the idea of having it as part of the deployment procedure. Though I think maybe it's better to run during the Docker build, and so we have the JSON specs already ready with the Docker image? 

This way should also be the easiest to implement, since it is similar to how admin panel's frontend assets are built today.

Also, for development environment, it would be nice to setup a worker that watches for file changes and invoke the generator, but this could be a separate PR.

InoMurko:
If you make it part of Docker build, you're forcing the docker deployment model. Is that intended?

unnawut:
Ah sorry I wasn't too clear. We're moving to a distillery release being prepared by @sirn, but in the meanwhile the current build setup relies on Docker. 

So if I understand correctly, using a distillery pre start hook now means this feature is dependent on when our distillery release is ready. Hence the suggestion to align with our current docker build process for this PR.

Thoughts? @sirn @T-Dnzt

sirn:
In the upcoming weeks, we'll be switching our build to use distillery and produce Docker image from distillery artifacts. Right now it could be a step in CI or Docker. I'll convert it to boot hook in the  branch.

Distillery branch is currently waiting for #240 to finalize (need to test with staging data, in which I got distracted with few other stuff) so I hope to get it done by the beginning of next week.

PR #239, opened 2018-06-12
Fix `Account.put_categories/3` to support both atom and string attribute keys

unnawut:
Issue/Task Number: T392

# Overview

This PR fixes the issue that `/account.update` does not update account categories even though category_ids are passed.

# Changes

- Update `Account.put_categories/3` to support both string and atom keys in attributes

# Usage

Try update account categories by passing a different `category_ids` during `/account.update`


jarindr:
question, why do we need to accept both atom string and array?

unnawut:
> question, why do we need to accept both atom string and array?

In practice with what we use now, we can expect the inputs to have string keys.

But `Ecto.Changeset.cast/4`, which we almost always use for creating a changeset, [allows both string keys and atom keys](https://github.com/elixir-ecto/ecto/blob/v2.2.10/lib/ecto/changeset.ex#L494-L543), so I prefer to keep it consistent.

Issue #232, opened 2018-06-08
Drop libsodium in favor for alternatives


sirn:
Currently we're using libsodium for metadata encryption in the eWallet DB and Local Ledger DB. Unfortunately, as libsodium (via enacl) is a NIF and I couldn't find a way to bundle libsodium along with release, this makes distributing eWallet binaries for supported platforms a bit harder than I'd hope (e.g. no cross compilation, for one thing). 

As libsodium is only used for metadata encryption via Cloak (hopefully not very important) it might more more feasible to switch to AES256-GCM instead. I'd love to hear what you think @thibault @unnawut. Also looping in @fred.

T-Dnzt:
Sounds good to me, I don't see any reason not to switch if it makes it easier to distribute binaries. Let's make it happen!

sirn:
@thibault Oops! 🤦‍♂️ You have my apologize. Very sorry about this.

sirn:
Implemented as of #240

PR #240, opened 2018-06-12
Upgrade Cloak to 0.7.0 for proper Umbrella support

sirn:
# Overview

Upgrade Cloak to **0.7.0-alpha.2** for proper Umbrella support, otherwise multi-keys support would remain broken as Cloak would attempt to use the key from the process that was first started. Cloak 0.7.0 introduced per-application "vaults" that allowed multiple keys to be used.

Since Cloak changed its cipher API, the old `Salty.SecretBox.Cloak` is no longer supported under 0.7.0. I've switched over to AES-GCM (see #232) as a new cipher going forward. The migration config will be removed in the [distillery-release](https://github.com/omisego/ewallet/tree/distillery-release) branch.

# Changes

- Upgrade Cloak to `0.7.0-alpha`
- Perform the [migration steps](https://hexdocs.pm/cloak/0.7.0-alpha.2/0.6.x_to_0.7.x.html#install-cloak-0-7) as outlined in Cloak documentation, including:
  - Create per-application vaults.
  - Changed `EncryptedMapField` to `*.Encrypted.Map` fields.
  - Drop `encryption_version` columns.
- Create a [cloak-0.7.0](https://github.com/omisego/salty/tree/cloak-0.7.0/) branch in Salty to provide migrate path to AES-GCM.

# Implementation Details

In Cloak version 0.6 and lower, only single instance of cloak application could be started at a time, which make using multiple encryption keys impossible. Our requirement of using different keys for eWallet DB and ledger DB is due to regulatory requirement in some territory.

Cloak introduced support for Umbrella in 0.7.0 which allowed multiple keys to be used in an application-level encryption module named `Vault`. 

# Usage

Everything should work the same, except now with `AES.GCM.V1` tag rather than `SBX`.

# Impact

Need to migrate existing key using:

1. Run `mix do deps.get, compile`
2. Run `mix ecto.migrate`
3. Run `mix omg.migrate.encryption`

For production, old key must be specified in the following environments:

* `RETIRED_EWALLET_SECRET_KEY`
* `RETIRED_LOCAL_LEDGER_SECRET_KEY`

For the new key, the following environments are used:

* `EWALLET_SECRET_KEY`
* `LOCAL_LEDGER_SECRET_KEY`

Keys may be generated using:

```
$ iex
iex> 32 |> :crypto.strong_rand_bytes() |> Base.encode64()
```


sirn:
I haven't really tested migrating old data to AES.GCM. Will do tomorrow.

docs/setup/env.md
```diff
@@ -4,73 +4,73 @@
 
 Below are the general environment variables needed for the eWallet to run smoothly.
 
-- `MIX_ENV`: Environment in which the application is being ran. `prod` for production.
-- `BASE_URL`: The base to use when building URLs.
-- `PORT`: The port that the application listens on.
-- `EWALLET_SECRET_KEY`: Encryption key used to encrypt some data in the database.
-- `LOCAL_LEDGER_SECRET_KEY`: Encryption key used to encrypt some data in the database.
+-   `MIX_ENV`: Environment in which the application is being ran. `prod` for production.
```
sirn: Weird spacing brought to you by [prettier](https://github.com/prettier/prettier).

sirn:
Just a heads-up, Cloak is [assuming numeric ID](https://github.com/danielberkompas/cloak/blob/eadb6a163acb2ca3039ea1162f21c8ed20252adc/lib/cloak/migrator.ex#L15) for its migrator so it looks like I'll have to patch it or write a custom migrator (probably the latter and submit patches later). Will work on it tomorrow.

sirn:
@unnawut @T-Dnzt I've added `omg.migrate.encryption` and would love a review look at it (probably another testing with staging data, although I've already tested locally.)

README.md
```diff
@@ -1,67 +1,67 @@
-OmiseGO eWallet
-===============
+# OmiseGO eWallet
```
sirn: Again, brought to you by [prettier](https://github.com/prettier/prettier). ¯\\\_(ツ)\_/¯

PR #242, opened 2018-06-13
Move all server endpoints from the eWallet API to the Admin API

T-Dnzt:
Issue/Task Number: T386

# Overview

This PR moves all the eWallet API sever endpoints to the Admin API. Feel free to ask questions if anything clear, the only changes I made other than moving things around are described below.

Note that for now, some endpoints will not work properly since they rely on a user being logged in. This will be fixed in a different PR to allow provider auth to work in those cases.

# Changes

- Move server endpoints
- Change the `OMGServer` auth scheme to `OMGProvider`
- Make the Admin API websocket-capable (channels are currently duplicated, which is a good temporary solution. We can revisit later to extract them to another sub application maybe.)
- Make endpoint path more consistent 
  - Rename all `list_` to `get_`
  - Move `/admin.upload_avatar` to `/me.upload_avatar`
  - Move `/logout` to `/me.logout`
  - Rename `/password.reset` to `/admin.reset_password`
  - Rename `/password.update` to `/admin.update_password`
  - Rename `/get_settings` to `/settings.all`
- Updated Swagger!


apps/ewallet/config/dev.exs
```diff
@@ -1,9 +1,19 @@
 use Mix.Config
 
+endpoints = []
+
+endpoints =
```
unnawut: Can you put some explanation as comment about why we need to do `Code.ensure_compiled?/1` and the conditioning?

apps/ewallet/lib/ewallet/fetchers/uuid_fetcher.ex
```diff
@@ -6,8 +6,9 @@ defmodule EWallet.UUIDFetcher do
 
   # Add more mappings if you wish to use this fetcher for other schemas.
   @mappings %{
-    "account_id" => Account,
-    "user_id" => User
+    "account_id" => {Account, :id, "account_uuid"},
```
unnawut: :mindblown: 🥇

apps/ewallet/lib/ewallet/web/v1/event_handlers/event.ex
```diff
@@ -21,15 +21,21 @@ defmodule EWallet.Web.V1.Event do
...
-          event,
-          payload
-        )
+        broadcast_if_compiled(endpoint, topic, event, payload)
       end)
     end)
   end
 
+  defp broadcast_if_compiled(endpoint, topic, event, payload) do
+    if Code.ensure_compiled?(endpoint) do
```
unnawut: What cases would make this not compiled?

apps/admin_api/lib/admin_api/v1/router.ex
```diff
@@ -91,25 +104,28 @@ defmodule AdminAPI.V1.Router do
...
     # Self endpoints (operations on the currently authenticated user)
     post("/me.get", SelfController, :get)
-    post("/me.get_account", SelfController, :get_account)
     post("/me.get_accounts", SelfController, :get_accounts)
+    post("/me.get_account", SelfController, :get_account)
     post("/me.update", SelfController, :update)
+    post("/me.upload_avatar", SelfController, :upload_avatar)
 
-    post("/logout", AuthController, :logout)
+    post("/admin.logout", AdminAuthController, :logout)
```
unnawut: Why not `/me.logout` to be consistent with routes above?

apps/admin_api/lib/admin_api/v1/router.ex
```diff
@@ -91,25 +104,28 @@ defmodule AdminAPI.V1.Router do
...
   end
 
   # Public endpoints (still protected by API key)
   scope "/", AdminAPI.V1 do
     pipe_through([:api, :client_api])
 
-    post("/login", AuthController, :login)
+    post("/admin.login", AdminAuthController, :login)
```
unnawut: And `/me.login`?

apps/admin_api/lib/admin_api/v1/controllers/transaction_controller.ex
```diff
@@ -76,6 +108,49 @@ defmodule AdminAPI.V1.TransactionController do
 
   def create(conn, _), do: handle_error(conn, :invalid_parameter)
 
+  defp contains_illegal_from_and_to?(terms) do
```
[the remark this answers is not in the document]
T-Dnzt: Definitely, let's add it to the list of things to change in the transaction endpoint refactoring! Don't want to do it here as it involves to move things around too much for my taste (and this is already big enough :D)

apps/ewallet/config/dev.exs
T-Dnzt: Sure, adding this:

```
# We use 'Code.ensure_compiled?' to ensure the endpoints are loaded.
# When running tests from inside the Admin API/eWallet API/eWallet sub apps,
# those endpoint modules are not available and therefore should not be added to
# the list of endpoints that needs to be notified. If they are and the
# Event emitter tries to call them, it will result in a 500 error.
```

This could probably be improved tho, by maybe extracting it to some functions? I wasn't sure where to put them tho, so just went with the copy approach. Open to suggestions!

apps/ewallet/lib/ewallet/web/v1/event_handlers/event.ex
T-Dnzt: Oh I forgot I actually moved the check here! You can discard my comment above, we can remove the `ensure_compiled?` from the config files. The explanation is however the same one:

```
# We use 'Code.ensure_compiled?' to ensure the endpoints are loaded.
# When running tests from inside the Admin API/eWallet API/eWallet sub apps,
# those endpoint modules are not available and therefore should not be added to
# the list of endpoints that needs to be notified. If they are and the
# Event emitter tries to call them, it will result in a 500 error.
```

apps/admin_api/lib/admin_api/v1/router.ex
T-Dnzt: Good point, changing it.

apps/admin_api/lib/admin_api/v1/router.ex
T-Dnzt: Since this is not an authenticated call, it doesn't fall in the same category as all the other `/me.` in which we can expect to have a user assigned in `conn`.

apps/ewallet/lib/ewallet/fetchers/uuid_fetcher.ex
T-Dnzt: 🕺

PR #243, opened 2018-06-14
Internal exchange mechanism

unnawut:
Issue/Task Number: T351

# Overview

Implements the functionality that allows cross-token exchanges and transactions.

Note that this is mostly behind the scene code. Integrating the exchange to the API endpoints is out of scope of this PR.

# Changes

- Added a migration that updates the `transfer` table schema to support exchanges
- Added migrations that updates inconsistencies in the table schemas
  - Rename ewallet_db's `transfer` to `transaction`
  - Swap names between local_ledger_db's `transaction` and `entry`
  - Uniquify id columns
- Added a migration and new schema for `EWalletDB.ExchangePair`
- Added `EWallet.Exchange` module to handle exchange-related calculation and validation
- Added tests for the whole `EWallet.TransactionFormatter` module
- Updated controllers and gates to support `from_amount`, `from_token`, `to_amount` and `to_token` attributes (except `EWallet.TransactionGate` which would impact endpoint request format)
- Updated `EWallet.TransferGate` to support cross-token transactions
- Updated `EWallet.TransactionFormatter` to support cross-token transactions
- Updated `LocalLedger.Transaction.Validator` to support cross-token transactions
- Updated `EWallet.TransactionGate` to validate attributes with `EWallet.Exchange.validate/4` before proceeding

# Implementation Details

Implemented according to [OIP-2](https://github.com/omisego/OIP/issues/2).

# Usage

The endpoint request/response formats have not been updated yet due to potential conflicts with other PRs. But you can see the working exchange and how it can be used in these tests:

- `EWalletDB.ExchangePairTest`
- `EWallet.ExchangeTest`
- `LocalLedger.Transaction.ValidatorTest`
- `EWallet.TransactionFormatterTest`

# Impact

New DB migrations. Run `mix ecto.migrate` after deploying this PR.


apps/local_ledger_db/lib/local_ledger_db/entry.ex
```diff
@@ -1,133 +1,166 @@
...
+    |> where([e], e.inserted_at <= ^upto)
   end
 
-  defp do_insert(changeset, opts) do
-    case Repo.insert(changeset, opts) do
-      {:ok, entry} ->
-        entry.idempotency_token
-        |> get_by_idempotency_token()
-        |> handle_retrieval_result()
+  defp build_sum_query(address, type, %{token_id: :all}) do
```
T-Dnzt: I was like "where is this all coming from?", then I remember we swapped them :D

apps/ewallet/lib/ewallet/exchange/exchange.ex
```diff
@@ -0,0 +1,154 @@
...
+  otherwise returns `false`.
+
+  For cross-token transactions, it returns `true` if the amounts match the rate
+  of the exchange pair, otherwise returns `false`.
+  """
+  @spec validate(non_neg_or_nil(), %Token{}, non_neg_or_nil(), %Token{}) ::
+          {:ok, Calculation.t()} | {:error, atom()}
+
+  # Same-token: valid if `from_amount` and `to_amount` to be equal
+  def validate(from_amount, %{uuid: from_token} = token, to_amount, %{uuid: to_token})
```
T-Dnzt: If I did this in one of my PR, you'd probably tell me to call it `from_token_uuid` instead of just `from_token`, just to clarify that we're manipulating an ID and not a token. :)

unnawut: Yeah I was compromising this with `mix format` trying to bump it to multiple lines and they look unreadable as hell to me 🤦‍♂️

apps/local_ledger_db/lib/local_ledger_db/entry.ex
unnawut: Yup. No code changes to these files. Only renames.

apps/local_ledger/lib/local_ledger/transaction/validator.ex
```diff
@@ -0,0 +1,100 @@
...
+    end
+  end
+
+  defp get_identical_addresses(entries) do
+    {debits, credits} = split_debit_credit(entries)
+
+    debit_addresses = extract_addresses(debits)
+    credit_addresses = extract_addresses(credits)
+
+    intersect(debit_addresses, credit_addresses)
```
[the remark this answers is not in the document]
unnawut: This is was your code 😅

apps/ewallet/test/ewallet/exchange/exchange_test.exs
```diff
@@ -0,0 +1,315 @@
...
+    test "returns {:error, :exchange_pair_not_found} if the exchange pair is not found",
+         context do
+      {res, code} = Exchange.get_rate(context.omg, context.eth)
+
+      assert res == :error
+      assert code == :exchange_pair_not_found
+    end
+  end
+
+  describe "get_rate/2 for tokens with diiferent subunit_to_unit values" do
```
mederic-p: typo diiferent :)

apps/ewallet/lib/ewallet/exchange/exchange.ex
unnawut: Wait... it doesn't do that anymore. What the hell.. 🤦‍♂️ 🤦‍♂️ 

Pushing the code now.

apps/local_ledger/lib/local_ledger/transaction/validator.ex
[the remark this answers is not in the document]
T-Dnzt: Lol, I'm impressing myself 😂

apps/ewallet_db/lib/ewallet_db/exchange_pair.ex
```diff
@@ -0,0 +1,171 @@
...
+  alias EWalletDB.{Repo, Token}
+
+  @primary_key {:uuid, UUID, autogenerate: true}
+
+  schema "exchange_pair" do
+    external_id(prefix: "exg_")
+
+    field(:name, :string)
+    field(:rate, :float)
+    field(:reversible, :boolean)
```
sirn: A little bit torn about handling `reversible`, can we not have an explicit reverse from/to pair, as in `from: "AAA", to: "BBB"` and `from: "BBB", to: "AAA"` as a separate exchange pairs? I think it might be simpler, logic-wise. Thoughts @T-Dnzt?

apps/ewallet/lib/ewallet/exchange/exchange.ex
```diff
@@ -0,0 +1,154 @@
...
+  """
+  @spec calculate(non_neg_or_nil(), %Token{}, non_neg_or_nil(), %Token{}) ::
+          {:ok, Calculation.t()} | {:error, atom()} | {:error, atom(), String.t()}
+  # Returns an :invalid_parameter error if both `from_amount` and `to_amount` are missing
+  def calculate(nil, _, nil, _) do
+    {:error, :invalid_parameter, "an exchange requires from amount, to amount, or both"}
+  end
+
+  # Returns an :invalid_parameter error if both `from_amount` and `to_amount` are provided
+  def calculate(from_amount, _, to_amount, _) when from_amount != nil and to_amount != nil do
```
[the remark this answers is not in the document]
unnawut: This is the current code:

```ex
def calculate(nil, _, nil, _) do
  # Returns error. At least one amount is needed.
end

def calculate(from_amount, _, to_amount, _) when from_amount != nil and to_amount != nil do
  # Returns error. Both amounts cannot be provided together.
end

def calculate(from_amount, %{uuid: same_uuid} = token, to_amount, %{uuid: same_uuid}) do
  # Return value for same token transaction
end

def calculate(nil, from_token, to_amount, to_token) do
  # Return value for cross token transactions, when from_amount is not provided.
end

def calculate(from_amount, from_token, nil, to_token) do
  # Return value for cross token transactions, when to_amount is not provided.
end
```

To remove this, I need to update the function guards above to:

```ex
def calculate(nil, _, nil, _) do
  # Returns error. At least one amount is needed.
end

def calculate(from_amount, %{uuid: same_uuid} = token, to_amount, %{uuid: same_uuid}) when from_amount == nil or to_amount == nil do
  # Return value for same token transaction
end

def calculate(nil, from_token, to_amount, to_token) when to_amount != nil do
  # Return value for cross token transactions, when from_amount is not provided.
end

def calculate(from_amount, from_token, nil, to_token) when from_amount != nil do
  # Return value for cross token transactions, when to_amount is not provided.
end

def calculate(_, _, _, _) do
  # Returns error. Both amounts cannot be provided together.
end
```

... which I feel the latter is more complex/confusing.

Unless of course if you have other suggestions to reduce the logic! 🙇

apps/ewallet_db/lib/ewallet_db/exchange_pair.ex
unnawut: I had reversible because I found that a few libraries in other languages do support reversible rates. But I'm ok to drop reversible support. It helps to reduce overall complexity like sirn said. Final decision for you @T-Dnzt.

T-Dnzt: Let's make it explicit by creating a separate exchange pair then. But can we keep the reversible parameter in the endpoints and automatically create that second separate pair from the API? Makes things faster than having to create 2 pairs every time :)

mederic-p: Oh, if we do this then there is no need to add the `reversible` field to the JSON

unnawut: @T-Dnzt If we go with auto-create the opposite pair,`reversible` sounds like an attribute that gets attached to the direct pair.

I propose renaming it to `create_opposite_pair: true/false` or something similar?

T-Dnzt: Sure! Just has to be in the API anyway, probably not in this PR? 🤔

unnawut: Yes can take it as a separate PR. But to make sure this PR is compatible with the direction above, I will remove reversible from `ExchangePair.cast()` and default the value to `false`.

unnawut: Ok so I decided to remove the `reversible` field from the database instead. So we don't need to add another DB migration (along with index changes) later.

PR #244, opened 2018-06-14
/category.update to support updating account_ids

unnawut:
Issue/Task Number: T402

# Overview

This PR adds support for `/category.update` to update its associated accounts by passing `account_ids`

# Changes

- Implemented `Category.put_accounts/3` and attach it to `Category.changeset/2`
- Refactored attribute retrieval into `EWalletDB.Helpers.InputAttribute.get/2`

# Implementation Details

This is basically the same implementation as the existing [`Account.put_categories/3`](https://github.com/omisego/ewallet/blob/develop/apps/ewallet_db/lib/ewallet_db/account.ex#L102-L120) but the other way around.

Also since the string/atom attribute retrieval is now used in multiple places, I refactored it into `EWalletDB.Helpers.InputAttribute.get/2`.

# Usage

Make a request to `/category.update` with `{"account_ids": [], ...}`

### DIFF
--- a/apps/ewallet_db/lib/ewallet_db/account.ex
+++ b/apps/ewallet_db/lib/ewallet_db/account.ex
@@ -102,6 +102,7 @@ defmodule EWalletDB.Account do
     case InputAttribute.get(attrs, attr_name) do
       ids when is_list(ids) ->
         put_categories(changeset, ids)
+
       _ ->
         changeset
     end
--- a/apps/ewallet_db/lib/ewallet_db/account.ex
+++ b/apps/ewallet_db/lib/ewallet_db/account.ex
@@ -9,6 +9,7 @@ defmodule EWalletDB.Account do
   import EWalletDB.{AccountValidator, Helpers.Preloader}
   alias Ecto.{Multi, UUID}
   alias EWalletDB.{Repo, Account, APIKey, Category, Key, Membership, Token, Wallet}
+  alias EWalletDB.Helpers.InputAttribute
 
   @primary_key {:uuid, UUID, autogenerate: true}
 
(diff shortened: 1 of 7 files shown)


apps/ewallet_db/lib/ewallet_db/account.ex
```diff
@@ -97,28 +98,23 @@ defmodule EWalletDB.Account do
...
+      _ ->
+        changeset
+    end
   end
 
-  defp put_categories(changeset, category_ids) when is_list(category_ids) do
+  defp put_categories(changeset, category_ids) do
     # Associations need to be preloaded before updating
     changeset = Map.put(changeset, :data, Repo.preload(changeset.data, :categories))
     categories = Repo.all(from(c in Category, where: c.id in ^category_ids))
```
T-Dnzt: I think we could use the `UUIDFetcher` for this, to make it more reusable throughout controllers. Good enough for now however, since this is a peculiar case. We might want to update it later :)

apps/ewallet_db/lib/ewallet_db/category.ex
```diff
@@ -24,7 +25,8 @@ defmodule EWalletDB.Category do
       :accounts,
       Account,
       join_through: "account_category",
-      join_keys: [category_uuid: :uuid, account_uuid: :uuid]
+      join_keys: [category_uuid: :uuid, account_uuid: :uuid],
+      on_replace: :delete
```
T-Dnzt: 👏

apps/ewallet_db/lib/ewallet_db/helpers/input_attribute.ex
```diff
@@ -0,0 +1,19 @@
+defmodule EWalletDB.Helpers.InputAttribute do
```
T-Dnzt: 👍

apps/ewallet_db/lib/ewallet_db/category.ex
unnawut: Ecto nicely informed me to do this:

```
     ** (RuntimeError) you are attempting to change relation :accounts of
     EWalletDB.Category, but there is missing data.

     If you are attempting to update an existing entry, please make sure
     you include the entry primary key (ID) alongside the data.

     If you have a relationship with many children, at least the same N
     children must be given on update. By default it is not possible to
     orphan embed nor associated records, attempting to do so results in
     this error message.

     If you don't desire the current behavior or if you are using embeds
     without a primary key, it is possible to change this behaviour by
     setting `:on_replace` when defining the relation. See `Ecto.Changeset`'s
     section on related data for more info.
```

PR #254, opened 2018-06-19
Fix channels in Admin/Client API

T-Dnzt:
Issue/Task Number: A

# Overview

Forgot to update the channels and the tests weren't failing due to their setup/structure

# Changes

- Remove channel authorization in AdminAPI
- Remove provider channel authorization from ClientAPI


T-Dnzt:
@unnawut not in the admin API (they should all still be in the eWallet API channels). We will need to add them back once we've figured out the permissions for the controllers (should be usable in the channels as well). Make sense?

PR #256, opened 2018-06-19
Update mint refine, format number

jarindr:
Issue/Task Number: - 

# Overview
Format numbers to `localString` with commas and update wallet when mint token.
this branch head is from #252 so some code is duplicate from #252 (merge #252 first)

# Changes
- Format numbers with commas.
- Update wallet when mint token successfully.
- Update wallet when create transaction successfully.
- ~~Add pooling functionality for `walletProvider`~~
- Add checkmark color for more visibility.
- Remove ability to fetch only once for wallet
# Implementation Details
- When we mint token, the token balance on master ( primary ) should be updated so we call `getWallets` of that account to update all accounts.
- Same as mint token, did the same as create transaction.
- Format numbers with comma with `toLocalString()` native javascript.

# Usage

- Look at numbers, it should contain comma.
- Try mint token and see the balance on primary wallet.

# Impact

Deploy as usual.

screen shot 2018-06-20 at 1 15 03 am


apps/admin_panel/assets/src/omg-wallet/walletProvider.js
```diff
@@ -8,12 +8,18 @@ class WalletProvider extends Component {
     render: PropTypes.func,
     walletAddress: PropTypes.string,
     getWalletById: PropTypes.func,
-    wallet: PropTypes.object
+    wallet: PropTypes.object,
+    pooling: PropTypes.bool
   }
 
   componentDidMount = () => {
     if (!this.props.wallet) {
```
T-Dnzt: Would there be a case when the `componentDidMount` is called and `this.props.wallet` is already set to something?

jarindr: @T-Dnzt yes, if you came from wallets page ( witch will fetch multiple wallets ) this case wallet will already exists and will not be fetch again.

T-Dnzt: So if that's the case, you wouldn't set up the auto-reload every 5 seconds, right? Isn't that something we want no matter what?

T-Dnzt: Oh you just fixed it, cool.

apps/admin_panel/assets/src/websocket/connector.js
```diff
@@ -0,0 +1,3 @@
+import Socket from 'phoenix-socket'
+const socket = new Socket('wss://localhost:4000/api/socket', { params: { userToken: '123' } })
```
T-Dnzt: Should probably be `/api/admin/socket` btw. Not sure why it's in this PR tho :D

jarindr: mistake lol it should be in another PR

PR #259, opened 2018-06-20
Allow transaction_request to be created with only token_id and type

T-Dnzt:
Issue/Task Number: T429

# Overview

Only `type` and `token_id` should be required to create transaction requests. However, currently some functions expect to see `correlation_id`, `amount` and `address` in the arguments. This PR removes those unnecessary checks.

# Changes

- Modify the min params test for transaction request gate.
- Remove unneeded params check in `TransactionRequestGate` functions


T-Dnzt:
@unnawut 😂 just fixing bugs reported by @mederic-p (a.k.a the Bug Finder) as soon as he finds them.

PR #261, opened 2018-06-20
Fix consumptions unneeded params

T-Dnzt:
Issue/Task Number: T433

# Overview

This PR removes unneeded params check from transaction request consumption and fix a bug returning a 500 when the amount used was the one coming from the transaction request.

# Changes

- Remove `address` as a mandatory param when consuming as a user
- Fix returned value when validating the amount of request/consumption

### DIFF
--- a/apps/ewallet/test/ewallet/gates/transaction_consumption_consumer_gate_test.exs
+++ b/apps/ewallet/test/ewallet/gates/transaction_consumption_consumer_gate_test.exs
@@ -414,31 +414,31 @@ defmodule EWallet.TransactionConsumptionConsumerGateTest do
          meta do
       initialize_wallet(meta.sender_wallet, 200_000, meta.token)
 
-      {res, consumption1} =
+      {res, consumption_1} =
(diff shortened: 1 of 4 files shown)


apps/ewallet/test/ewallet/gates/transaction_consumption_consumer_gate_test.exs
```diff
@@ -414,6 +414,36 @@ defmodule EWallet.TransactionConsumptionConsumerGateTest do
...
+          "formatted_transaction_request_id" => meta.request.id,
+          "idempotency_token" => "123"
+        })
+
+      assert res == :ok
+      assert %TransactionConsumption{} = consumption
+      assert consumption.transaction_request_uuid == meta.request.uuid
+      assert consumption.amount == meta.request.amount
+      assert consumption.wallet_address == meta.sender_wallet.address
+    end
```
unnawut: Since the test case says "(and is idempotent)", can you also assert that `consumption1.transaction_uuid` and `consumption2.transaction_uuid` is the same to guarantee idempotency?

Also applies to the test case below.

PR #265, opened 2018-06-22
Allow API keys to be disabled/enabled

T-Dnzt:
Issue/Task Number: T323

# Overview

Add an `/api_key.update` endpoint to allow admin to enable/disable keys.

# Changes

- Add `/api_key.update`
- Update OpenAPI spec

# Usage

Call `/api_key.update` with an `id` and `expired` set to `true` or `false`.

See OpenAPI spec for more info.

### DIFF
--- a/apps/ewallet_db/lib/ewallet_db/api_key.ex
+++ b/apps/ewallet_db/lib/ewallet_db/api_key.ex
@@ -28,7 +28,7 @@ defmodule EWalletDB.APIKey do
       type: UUID
     )
 
-    field(:expired, :boolean)
+    field(:expired, :boolean, default: false)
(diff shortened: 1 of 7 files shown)


apps/ewallet_db/lib/ewallet_db/api_key.ex
```diff
@@ -41,6 +41,12 @@ defmodule EWalletDB.APIKey do
     |> assoc_constraint(:account)
   end
 
+  defp update_changeset(%APIKey{} = key, attrs) do
```
unnawut: Yes! Thank you for letting updates only on specific fields! 🎉

PR #268, opened 2018-06-25
Admin transaction requests

T-Dnzt:
Issue/Task Number: T365

# Overview

This PR adds the missing endpoints to deal with transaction requests and consumptions. They can also be used to do long-polling of the pending consumptions for accounts/users/wallets or transaction requests (by sending `search_terms: {status: "pending"}`.

# Changes

- Added `/transaction_request.all`
- Added `/account.get_transaction_consumptions`
- Added `/user.get_transaction_consumptions`
- Added `/wallet.get_transaction_consumptions`
- Added `/transaction_request.get_transaction_consumptions`
- Added `/transaction_consumptions.all`
- Added `/transaction_consumptions.get`

To get only pending ones, the four endpoints using `/entity.get_transaction_consumptions` need the param `search_terms` to be passed with `{status: "pending"}`.

# Implementation Details

Due to the way the `SearchParser` is implemented, it is not possible to reliably use both `search_term` and filter per account/user/whatever (because of the way the queries are built with `OR` in the search parser, can't find a way to have `(X) AND (A OR B)` without using Ecto Dynamic Queries and basically rewriting the whole Searcher and updating all the controllers. Instead, I just prevent the use of `search_term`, so when using those endpoints, only `search_terms` will work.

# Usage

See OpenAPI spec.

### DIFF
--- a/apps/admin_api/priv/spec.yaml
+++ b/apps/admin_api/priv/spec.yaml
@@ -904,6 +904,23 @@ paths:
   ############################
   #   TRANSACTION REQUESTS   #
   ############################
+  # Endpoint to list all transaction requests
+  /transaction_requests.all:
+    post:
+      tags:
+        - TransactionRequest
+      summary: Get the list of transaction requests
+      operationId: transaction_requests_all
+      security:
+        - ProviderAuth: []
+        - AdminAuth: []
+      requestBody:
+        $ref: '#/components/requestBodies/TransactionRequestAllBody'
+      responses:
+        '200':
+          $ref: "#/components/responses/TransactionRequestsResponse"
+        '500':
+          $ref: "#/components/responses/InternalServerError"
   # Endpoint to create a transaction request
   /transaction_request.create:
     post:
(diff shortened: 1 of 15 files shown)


apps/admin_api/priv/spec.yaml
```diff
@@ -955,6 +972,109 @@ paths:
...
+        - AdminAuth: []
+      requestBody:
+        $ref: '#/components/requestBodies/TransactionConsumptionAllForTransactionRequestBody'
+      responses:
+        '200':
+          $ref: "#/components/responses/TransactionConsumptionsResponse"
+        '500':
+          $ref: "#/components/responses/InternalServerError"
+  # Endpoint to list all consumptions
+  /transaction_consumptions.all:
```
unnawut: no `s`

apps/admin_api/priv/spec.yaml
```diff
@@ -955,6 +972,109 @@ paths:
...
+    post:
+      tags:
+        - TransactionConsumption
+      summary: Get the list of transaction consumptions
+      operationId: transaction_consumptions_all
```
unnawut: no `s`

T-Dnzt:
@unnawut Ready for another look!

PR #270, opened 2018-06-26
Improve amounts displayed in insufficient funds error

mederic-p:
Issue/Task Number: T449

# Overview

This PR improves the display amount in `insufficient_funds` errors.
There was 2 issues in the current implementation:
- trailing zeros (ie: 10000.0)
- decimals limited to 1 (ie: 1234.5), so we were missing accuracy if the subunit_to_unit was bigger than 10

# Changes

- Created a `AmountFormatter` that takes care of formatting amount for display
- The number of decimals is now dynamic depending on the subunit_to_unit given
- The string is formatted so if it ends with '.0' it's truncated

### DIFF
--- a/apps/ewallet/lib/ewallet/formatters/amount_formatter.ex
+++ b/apps/ewallet/lib/ewallet/formatters/amount_formatter.ex
@@ -3,18 +3,19 @@ defmodule EWallet.AmountFormatter do
   A string formatter for amounts
   """
   def format(amount, subunit_to_unit) do
-    float_amount = amount / subunit_to_unit
-
-    float_amount
-    |> float_to_binary(to_decimals(subunit_to_unit))
+    subunit_to_unit
+    |> to_decimals()
+    |> float_to_binary(amount / subunit_to_unit)
     |> String.replace_trailing(".0", "")
   end
 
-  defp float_to_binary(value, decimals) do
+  defp float_to_binary(decimals, value) do
     :erlang.float_to_binary(value, [:compact, {:decimals, decimals}])
   end
 
   defp to_decimals(subunit_to_unit) do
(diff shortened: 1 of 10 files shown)


apps/ewallet/lib/ewallet/formatters/amount_formatter.ex
```diff
@@ -0,0 +1,20 @@
...
+    |> float_to_binary(to_decimals(subunit_to_unit))
+    |> String.replace_trailing(".0", "")
+  end
+
+  defp float_to_binary(value, decimals) do
+    :erlang.float_to_binary(value, [:compact, {:decimals, decimals}])
+  end
+
+  defp to_decimals(subunit_to_unit) do
+    Kernel.trunc(:math.log10(subunit_to_unit))
```
T-Dnzt: Can't we use a pipeline here?

```
subunit_to_unit
|> :math.log10()
|> Kernel.trunc()
```

apps/ewallet/lib/ewallet/formatters/amount_formatter.ex
```diff
@@ -0,0 +1,20 @@
+defmodule EWallet.AmountFormatter do
+  @moduledoc """
+  A string formatter for amounts
+  """
+  def format(amount, subunit_to_unit) do
+    float_amount = amount / subunit_to_unit
+
+    float_amount
+    |> float_to_binary(to_decimals(subunit_to_unit))
+    |> String.replace_trailing(".0", "")
```
T-Dnzt: If you swap `value` and `decimals` in `float_to_binary`, we can have something a bit cleaner:
```
subunit_to_unit
|> to_decimals()
|> float_to_binary(float_amount)
|> String.replace_trailing(".0", "")
```

PR #271, opened 2018-06-26
Rename constraint names to align with latest table schemas

unnawut:
Issue/Task Number: T330

# Overview

Fixes quite a few constraint names that were out of sync due to columns being renamed over time.

# Changes

- Add a new DB migration that aligns the constraint names with the latest table column names

# Implementation Details

I decided to use `execute()` with raw SQLs due to readability and not having to depend on column properties.

This `ALTER TABLE ... RENAME CONSTRAINT ...` is available since postgres 9.1 released in 2011 so it should be safe enough.

All constraints in LocalLedgerDB looks correct so there's no migration for that.

# Usage

Run `mix ecto.migrate`.

To check all the constraint names, run the following SQL:

```sql
SELECT
    tc.constraint_name,
    tc.table_name || '.' || kcu.column_name AS table_column
FROM 
    information_schema.table_constraints AS tc 
    JOIN information_schema.key_column_usage AS kcu
      ON tc.constraint_name = kcu.constraint_name
    JOIN information_schema.constraint_column_usage AS ccu
      ON ccu.constraint_name = tc.constraint_name;

-- Thanks to https://gist.github.com/PickledDragon/dd41f4e72b428175354d
```

# Impact

Run `mix ecto.migrate` after deploy.

### DIFF
--- a/apps/ewallet_db/priv/repo/migrations/20180626055050_update_constraint_names.exs
+++ b/apps/ewallet_db/priv/repo/migrations/20180626055050_update_constraint_names.exs
@@ -1,0 +1,45 @@
+defmodule EWalletDB.Repo.Migrations.UpdateConstraintNames do
+  use Ecto.Migration
+
+  @renames [
+    # {table, from_constraint, to_constraint},
+    {:account, "account_parent_id_fkey", "account_parent_uuid_fkey"},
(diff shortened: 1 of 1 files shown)


unnawut:
Changing this PR's base to `t330-constraint-names` to avoid conflicts with DB changes in #243.

unnawut:
Will merge after #243 is merged to avoid DB migration conflicts.

PR #273, opened 2018-06-26
Update address format to align with OIP-3

unnawut:
Issue/Task Number: T357

# Overview

This PR updates address format to align with [OIP-3](https://github.com/omisego/OIP/issues/3).

# Changes

- Add new Ecto type `EWalletDB.Types.WalletAddress`
- Change `EWalletDB.Wallet` to use `EWalletDB.Types.WalletAddress` instead of plain string
- `EWalletDB.Wallet` autogenerates the address if one is not given
- `EWalletDB.Wallet`'s address is immutable once set

# Implementation Details

- Mainly the added logic is in `EWalletDB.Types.WalletAddress` and implemented as a [custom Ecto type](https://hexdocs.pm/ecto/Ecto.Type.html).
- Address is saved in database with non-alphanumerics stripped so that it's consistent for searching and manipulating. Each provider can add non-alphanumerics separators or send them over unstripped as they wish.
- No migrations were added since the old addresses are still usable and its not ideal to change the addresses of existing wallets.

# Usage

Try `/account.create`, `/account.all`, `/account.get`, `/account.update`, etc.

# Impact

Only code changes.

### DIFF
--- a/apps/ewallet_db/lib/ewallet_db/wallet.ex
+++ b/apps/ewallet_db/lib/ewallet_db/wallet.ex
@@ -15,6 +15,8 @@ defmodule EWalletDB.Wallet do
   @primary "primary"
   @secondary "secondary"
 
+  @genesis_address "gnis000000000000"
+
   def genesis, do: @genesis
   def burn, do: @burn
   def primary, do: @primary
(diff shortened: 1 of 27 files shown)


apps/ewallet_db/lib/ewallet_db/wallet.ex
```diff
@@ -14,6 +15,8 @@ defmodule EWalletDB.Wallet do
   @primary "primary"
   @secondary "secondary"
 
+  @genesis_address "gnis000000000000"
```
T-Dnzt: We need the genesis address to be updated :)

unnawut:
@T-Dnzt Updated. Please check again.

PR #275, opened 2018-06-26
Account Authorization

T-Dnzt:
Issue/Task Number: T237

# Overview

This PR adds the actual permissions for the account controllers. 

# Changes

- Only send back accounts the current user/key is allowed to access
- Fix `SearchParser` and prevent its `or` queries from messing with the rest of the queries.
- Update `AccountPolicy` to handle recursive checks of permissions
- Add a bunch of functions to Account to navigate around the trees (get ancestors, descendants, etcc.)

# Implementation Details

Coming soon (maybe).


apps/ewallet_db/lib/ewallet_db/account.ex
```diff
@@ -339,6 +345,124 @@ defmodule EWalletDB.Account do
...
+    |> Enum.map(fn descendant -> descendant.id end)
+    |> Enum.member?(descendant_id)
+  end
+
+  @spec get_all_descendants_uuids(%Account{}) :: List.t()
+  def get_all_descendants_uuids(account) do
+    query_family_uuids(account, "
+      WITH RECURSIVE accounts_cte(uuid, parent_uuid, depth) AS (
+        SELECT ta.uuid, ta.parent_uuid, $1::INT AS depth
+        FROM account AS ta WHERE ta.uuid = $2
```
unnawut: Would be nice if we use full names instead of `ta`, `c`, etc. for these cases to help with readability

apps/ewallet/lib/ewallet/fetchers/account_fetcher.ex
```diff
@@ -0,0 +1,14 @@
+defmodule EWallet.AccountFetcher do
+  @moduledoc """
+  Module responsible for loading accounts.
+  """
+  alias EWalletDB.User
+
+  def get_highest_account(%{admin_user: admin_user}) do
```
unnawut: Because users can have memberships on separate descendant lines, it's possible that a user has multiple highest accounts. This probably need to handle multiple accounts case.

apps/ewallet_db/lib/ewallet_db/account.ex
```diff
@@ -339,6 +345,124 @@ defmodule EWalletDB.Account do
...
+      UNION ALL
+       SELECT c.uuid, c.parent_uuid, p.depth - 1 AS depth
+       FROM accounts_cte AS p, account AS c WHERE c.uuid = p.parent_uuid
+      )
+      SELECT uuid FROM accounts_cte AS a ORDER BY depth ASC
+      ")
+  end
+
+  defp query_family_uuids(account, query) do
```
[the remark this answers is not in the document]
T-Dnzt: I'm guessing you mean `get_family_uuids()`?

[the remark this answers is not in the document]
unnawut: Sorry... yes! That!

apps/ewallet_db/lib/ewallet_db/role.ex
```diff
@@ -25,15 +26,26 @@ defmodule EWalletDB.Role do
...
+    |> cast(attrs, [:priority, :name, :display_name])
+    |> validate_required([:name, :priority])
     |> unique_constraint(:name)
   end
 
   @doc """
   Creates a new role with the passed attributes.
   """
   def insert(attrs) do
+    last_role = Role |> Query.last(:inserted_at) |> Repo.one()
```
unnawut: Can we use `Ecto.Query.API.max(:priority)` instead? In case the last inserted happen to have a very high priority, so this one happens to get high priority by chance.

apps/ewallet_db/lib/ewallet_db/user.ex
```diff
@@ -448,14 +454,18 @@ defmodule EWalletDB.User do
     |> Repo.all()
   end
 
+  @spec get_membership_account_uuids(%User{}) :: List.t()
```
unnawut: It's using `Map.fetch!/2` so it could return `no_return()` too?

apps/ewallet_db/lib/ewallet_db/role.ex
```diff
@@ -25,15 +26,26 @@ defmodule EWalletDB.Role do
 
   defp changeset(%Role{} = key, attrs) do
     key
-    |> cast(attrs, [:name, :display_name])
-    |> validate_required(:name)
+    |> cast(attrs, [:priority, :name, :display_name])
+    |> validate_required([:name, :priority])
```
unnawut: Curious question: `:priority` can also be a negative integer right?

apps/ewallet/lib/ewallet/policies/account_policy.ex
```diff
@@ -3,34 +3,61 @@ defmodule EWallet.AccountPolicy do
...
+    |> do_authorize(account_id)
+  end
+
+  def do_authorize(memberships, account_id) do
+    # optimize this by allowing ancestors to be queried by account_id
+    # move this to another module.
+    with %Account{} = account <- Account.get(account_id) || {:error, :unauthorized},
+         ancestors <- Account.get_all_ancestors(account),
+         ancestors_uuids <- Enum.map(ancestors, fn ancestor -> ancestor.uuid end),
+         membership_accounts_uuids <-
```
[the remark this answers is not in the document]
T-Dnzt: Hmm, not sure if that optimizes it. `ancestors_uuids` is a list here, so inside the `any?`, we need another `any?` checking if the `ancestors_uuids` list contains the `membership.account_uuid`. 🤔

apps/ewallet_db/lib/ewallet_db/role.ex
T-Dnzt: Yup!

apps/ewallet_db/lib/ewallet_db/role.ex
```diff
@@ -35,15 +35,15 @@ defmodule EWalletDB.Role do
   Creates a new role with the passed attributes.
   """
   def insert(attrs) do
-    last_role = Role |> Query.last(:inserted_at) |> Repo.one()
+    highest_priority = Repo.one(from(r in Role, select: max(r.priority)))
```
unnawut: Was thinking of `lowest_priority` but `lowest_priority` could also mean lowest value, which has the high privileges 🤦‍♂️ Couldn't think of a better name at the moment.

PR #278, opened 2018-06-27
Fix null in error messages / Fix error description

mederic-p:
Issue/Task Number: T411

# Overview

This PR fixes the error messages and description returned in the JSON.

# Changes

- Replace the %{xxx} with the actual value in the description
- Add the error instead of `null` in messages

### DIFF
--- a/apps/ewallet/lib/ewallet/web/v1/error_handler.ex
+++ b/apps/ewallet/lib/ewallet/web/v1/error_handler.ex
@@ -315,9 +315,9 @@ defmodule EWallet.Web.V1.ErrorHandler do
   end
 
   defp stringify_errors(changeset, description) do
-    Enum.reduce(changeset.errors, description, fn {field, {description, _values}}, acc ->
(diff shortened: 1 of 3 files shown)


apps/ewallet/lib/ewallet/web/v1/error_handler.ex
```diff
@@ -315,9 +315,9 @@ defmodule EWallet.Web.V1.ErrorHandler do
   end
 
   defp stringify_errors(changeset, description) do
-    Enum.reduce(changeset.errors, description, fn {field, {description, _values}}, acc ->
+    Enum.reduce(changeset.errors, description, fn {field, {description, values}}, acc ->
       field = field |> stringify_field() |> replace_uuids()
-      acc <> " " <> field <> " " <> description <> "."
+      acc <> " " <> field <> " " <> replace_placeholders(description, values) <> "."
```
unnawut: We already have `build_template()` above that does something similar. Could you maybe use it and change it to `%{...}` format instead so we align with `Ecto.Changeset`'s errors?

PR #282, opened 2018-06-28
Add amount validation on transactions

mederic-p:
Issue/Task Number: 420

# Overview

This PR limits the maximum amount (from and to) that can be passed to a transaction.
Also fixes an issue with the current implementation of the `AmountFormatter` that was losing precision with big numbers

# Changes

- Limit `from_amount` and `to_amount` to be less than `100_000_000_000_000_000_000_000_000_000_000_000`

# Implementation Details

This constraint comes from a limitation of `swift` where we can't easily decode numbers with more than 38 digits.

### DIFF
--- a/apps/ewallet/lib/ewallet/formatters/amount_formatter.ex
+++ b/apps/ewallet/lib/ewallet/formatters/amount_formatter.ex
@@ -3,19 +3,10 @@ defmodule EWallet.AmountFormatter do
   A string formatter for amounts
   """
   def format(amount, subunit_to_unit) do
-    subunit_to_unit
-    |> to_decimals()
-    |> float_to_binary(amount / subunit_to_unit)
-    |> String.replace_trailing(".0", "")
-  end
+    Decimal.set_context(%Decimal.Context{Decimal.get_context() | precision: 38})
 
-  defp float_to_binary(decimals, value) do
-    :erlang.float_to_binary(value, [:compact, {:decimals, decimals}])
-  end
-
-  defp to_decimals(subunit_to_unit) do
-    subunit_to_unit
-    |> :math.log10()
-    |> Kernel.trunc()
+    amount
+    |> Decimal.div(subunit_to_unit)
+    |> Decimal.to_string(:normal)
   end
 end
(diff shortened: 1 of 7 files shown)


apps/ewallet/lib/ewallet/formatters/amount_formatter.ex
```diff
@@ -3,19 +3,10 @@ defmodule EWallet.AmountFormatter do
...
-    subunit_to_unit
-    |> to_decimals()
-    |> float_to_binary(amount / subunit_to_unit)
-    |> String.replace_trailing(".0", "")
-  end
-
-  defp float_to_binary(decimals, value) do
-    :erlang.float_to_binary(value, [:compact, {:decimals, decimals}])
-  end
+    Decimal.set_context(%Decimal.Context{Decimal.get_context() | precision: 38})
```
unnawut: `Decimal` is available because it's a dependency of `ecto` and `postgrex`.

Now that we're using it directly, can you add `{:decimal, "~> 1.0"}` to `apps/ewallet/mix.exs` so we explicitly require it?

unnawut: Also, can we refactor to this and move it to `EWallet.Application`?

```ex
# apps/ewallet/lib/ewallet/application.ex
defmodule EWallet.Application do
  # ..
  def start(_type, _args) do
    import Supervisor.Spec

    set_decimal_context()

    # ..
  end

  defp set_decimal_context do
    Decimal.get_context()
    |> Map.put(:precision, 38)
    |> Decimal.set_context()
  end
end
```

The refactor to pipes is for readability. The move to `Application` is because `Decimal.set_context()` is setting the context globally so it would be nice to set at application level rather than at specific functions.

T-Dnzt: Yup, good point.

T-Dnzt: Yeah, that's nice too 👍

mederic-p: Yeah I was wondering why it was available :)

PR #286, opened 2018-06-28
Prevents an exchange rate of 0 or below

unnawut:
Issue/Task Number: T454

# Overview

This PR prevents creating or updating an exchange pair with a rate of zero or below.

# Changes

- Update `ExchangePair`'s changeset to validate for a rate greater than 0
- Added tests for the controllers

# Implementation Details

Straightforward use of [`Ecto.Changeset.validate_number/3`](https://hexdocs.pm/ecto/Ecto.Changeset.html#validate_number/3)

# Usage

Try `/exchange_pair.create` or `/exchange_pair.update` with `rate` of 0 or below. They should return `invalid_parameter` error with the description explaining the validation error.

# Impact

Code changes only. No changes to the DB schema or API request/response format.

### DIFF
--- a/apps/admin_api/test/admin_api/v1/controllers/admin_auth/exchange_pair_controller_test.exs
+++ b/apps/admin_api/test/admin_api/v1/controllers/admin_auth/exchange_pair_controller_test.exs
@@ -98,6 +98,24 @@ defmodule AdminAPI.V1.AdminAuth.ExchangePairControllerTest do
       assert response["data"]["rate"] == 2.00
     end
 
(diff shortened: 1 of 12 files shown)


T-Dnzt:
@unnawut Could you also check and prevent an exchange from happening that would result in one of the amount being lower than the lowest integer (including `subunit_to_unit`)?

So if I send from_amount=null, to_amount=1 (as in 0.01 with subunit_to_unit=2) and rate between the tokens=2, from_amount should not be `0.5` and I should get an error instead?

unnawut:
@T-Dnzt Please check again

PR #288, opened 2018-06-29
Attempt to fix channel tests failures

T-Dnzt:
Issue/Task Number: T465

# Overview

 This PR contains attempts to fix the random failures for the channel tests.

### DIFF
--- a/apps/admin_api/test/support/channel_case.ex
+++ b/apps/admin_api/test/support/channel_case.ex
@@ -13,7 +13,7 @@ defmodule AdminAPI.ChannelCase do
   of the test unless the test case is marked as async.
   """
 
-  use ExUnit.CaseTemplate
+  use ExUnit.CaseTemplate, async: false
   alias Ecto.Adapters.SQL.Sandbox
 
   using do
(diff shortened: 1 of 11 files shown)


apps/admin_api/test/support/channel_case.ex
```diff
@@ -13,7 +13,7 @@ defmodule AdminAPI.ChannelCase do
   of the test unless the test case is marked as async.
   """
 
-  use ExUnit.CaseTemplate
+  use ExUnit.CaseTemplate, async: false
```
unnawut: Note the `defmacro __using__(_) do ...` in [`ExUnit.CaseTemplate.__using__/1`](https://github.com/elixir-lang/elixir/blob/v1.6.6/lib/ex_unit/lib/ex_unit/case_template.ex#L32-L45). It's ignoring all params.

But the 2nd level of `__using__/1` in there does proxy to [`use ExUnit.Case, unquote(opts)`](https://github.com/elixir-lang/elixir/blob/v1.6.6/lib/ex_unit/lib/ex_unit/case_template.ex#L50). 

So I think to make these test cases async, we need to update `AdminAPI.V1.AccountChannelTest`, `AdminAPI.V1.WalletChannelTest`, etc. and change `use AdminAPI.ChannelCase` to `use AdminAPI.ChannelCase, async: false` instead?

PR #291, opened 2018-07-02
Allow `create_opposite: true/false` on `/exchange_pair.create`

unnawut:
Issue/Task Number: T452

# Overview

Adds `create_opposite: true/false` to /exchange_pair.create to also create the opposite pair.

# Changes

Direct changes:
- Implement `sync_opposite: true/false` for `EWallet.ExchangePairGate.insert/1`
- Implement `sync_opposite: true/false` for `EWallet.ExchangePairGate.update/2`
- Implement `sync_opposite: true/false` for `EWallet.ExchangePairGate.delete/2`
- If the operations with sync_opposite fail midway, all data from that operation is rolled back.

Fixes:
- Validate that `from_token` and `to_token` are required 🤦‍♂️
- Validate that `from_token` and `to_token` are immutable after the pair is created
- Validate that `from_token` and `to_token` cannot be the same token

Side changes (required to make the above work):
- Update exchange_pair's (from_token, to_token, deleted_at) constraint to prevent duplicate pairs being allowed due to Postgres considering each `deleted_at: null` in the index as a unique value
- Update `EWalletDB.SchemaCase.test_update_prevents_changing/2` macro to support passing arbitary values so tokens can be passed.
- Add `EWalletDB.Validator.validate_different_values/3` to be used by ExchangePair to ensure a pair always has different tokens.
- Move `/exchange_pair.create`, `/exchange_pair.update` and /exchange_pair.delete` logic from the controller to the gate as they are now getting complex.
- Add a new `UUIDFetcher. load_record/3` guard to prevent querying with nil value
- Add typespecs to `EWalletDB.SoftDelete` functions

# Implementation Details

The `create_opposite: true/false` is handled at the gate and not at schema level because I treat this as a helper to create two exchange pairs at once, rather than a built in functionality of the exchange pair.

The field is named `create_opposite` rather than `reversed` to hint that a separate pair is created, and the reversal is not part of the direct pair. This is important as when one updates the rate of the direct pair, the opposite pair's rate will not be reflected automatically.

Also, the new validations are not required for this PR but since the code has been touched, I decided to add them in as well.

# Usage

Try `/exchange_pair.create` with param `create_opposite: true`. A list of two exchange pairs should be returned as the response.

# Impact

Code changes only. No DB migration. For `/exchange_pair.create` endpoint, `create_opposite` param is added to the request format, and the response is changed to a list of token objects instead of one token object.


apps/ewallet/test/ewallet/gates/exchange_pair_gate_test.exs
```diff
@@ -0,0 +1,80 @@
...
+    end
+  end
+
+  describe "update/2" do
+    test "updates an exchange pair" do
+      pair = insert(:exchange_pair)
+
+      {res, pair} =
+        ExchangePairGate.update(pair.id, %{
+          "name" => "Test pair updated",
```
T-Dnzt: I think we should have `update_opposite`as well :D

unnawut:
By the time test passes this branch becomes out of date with develop so I'll just rebase when it's ready to merge.

@T-Dnzt Ready for another review

apps/ewallet_db/priv/repo/migrations/20180704095243_update_unique_index_on_exchange_pair.exs
```diff
@@ -0,0 +1,24 @@
+defmodule EWalletDB.Repo.Migrations.UpdateUniqueIndexOnExchangePair do
+  use Ecto.Migration
+
+  def up do
+    execute ~s"DROP INDEX exchange_pair_from_token_uuid_to_token_uuid_deleted_at_index"
+
+    execute ~s"CREATE UNIQUE INDEX exchange_pair_from_token_uuid_to_token_uuid_deleted_at_not_null
```
T-Dnzt: You've given up on Ecto? 😅

T-Dnzt:
I was thinking, do we really want to let people decide the pair names? Shouldn't it be generated as `Tok1/tok2` by default?

apps/ewallet_db/priv/repo/migrations/20180704095243_update_unique_index_on_exchange_pair.exs
unnawut: Oh Ecto does support [partial indexes](https://hexdocs.pm/ecto/Ecto.Migration.html#index/3-partial-indexes)! Updating...

unnawut:
> I was thinking, do we really want to let people decide the pair names? Shouldn't it be generated as Tok1/tok2 by default?

Actually if we want to go with default naming, we can remove the customizable name altogether.

The name can be generated by the provider as they see fit using the exchange pair's data. We can also provide an auto-generated name, e.g. `token1.symbol <> "/" <> token2.symbol`.

We can always add the customizable name field back easily if there’s strong enough request/use case for it.

PR #292, opened 2018-07-03
Handle number string inputs

T-Dnzt:
Issue/Task Number: T479

# Overview

This PR adds support for string numbers for minting, transaction, requests and consumptions.


apps/admin_api/lib/admin_api/v1/controllers/token_controller.ex
```diff
@@ -82,15 +82,31 @@ defmodule AdminAPI.V1.TokenController do
       |> Map.put("account_uuid", Account.get_master_account().uuid)
       |> Token.insert()
 
-    case attrs["amount"] do
-      amount when is_number(amount) and amount > 0 ->
-        inserted_token
-        |> MintGate.mint_token(%{"amount" => amount})
-        |> respond_single(conn)
+    create_with_mint(attrs["amount"], inserted_token, conn)
```
unnawut: Can we validate for a valid amount and return error, even before token is inserted?

apps/ewallet/lib/ewallet/gates/transaction_request_gate.ex
```diff
@@ -181,24 +181,13 @@ defmodule EWallet.TransactionRequestGate do
...
-        else: attrs["allow_amount_override"]
-      )
+    require_confirmation = default_to_if_nil(attrs["require_confirmation"], false)
+    allow_amount_override = default_to_if_nil(attrs["allow_amount_override"], true)
 
     TransactionRequest.insert(%{
       type: attrs["type"],
       correlation_id: attrs["correlation_id"],
-      amount: attrs["amount"],
+      amount: get_integer_or_string_amount(attrs["amount"]),
```
unnawut: Can we do this in the `with` clause in `create()` instead?

PR #293, opened 2018-07-03
Allow cross-token transactions from/to exchange wallet

unnawut:
Issue/Task Number: T467

# Overview

This PR fixes the entries generation for cross-token transfers that involve `from` or `to` address being the exchange wallet itself.

# Changes

- Add support for cross-token transfers from/to an exchange wallet into `EWallet.TransactionFormatter`
- Add tests to validate the change in `TransactionFormatterTest` as well as transaction controller tests.
- Fill in some missing test cases between admin user and provider requests.

# Implementation Details

Because usual cross-token transfers require 4 debit/credit entries (2 for moving `from_token` into the exchange account, 2 for moving `to_token` out of the exchange account), so the existing `TransactionFormatter` generates a pair of debit/credit that goes to the same wallet address when an exchange wallet is also used as `from` or `to` address.

This is fixed by having the `TransactionFormatter` detects that if the `from`/`to` is the same address as the exchange wallet, it should generate only 1 pair of entries that moves the funds from the other wallet into the exchange wallet or vice versa.

# Usage

Try `/transaction.create` with `from` or `to` address being the same as the exchange wallet, or `/me.create_transaction` with `to` address being the same as the exchange wallet. The operations should be successful.

# Impact

Code changes only. No DB migrations or API request/response format changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,7 @@ Here are some resources if you want to learn more about how the eWallet works.
 -   [eWallet Entites](/docs/design/entities.md)
 -   [eWallet Components](/docs/design/components.md)
 -   [A closer look at wallets](/docs/design/wallets.md)
+-   [Transactions and entries](/docs/design/transactions_and_entries.md)
 
 # Contributing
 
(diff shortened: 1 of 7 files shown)


unnawut:
@T-Dnzt Done!

PR #299, opened 2018-07-05
Enable transaction tab, and Exchange rate functionality.

jarindr:
Issue/Task Number: 480-token-stats, 477-transaction-tab,475-exchange-with-rate,474-create-exchange-rate-modal

# Overview

Add functionality to view transaction, create rate for tokens and exchange with rate

# Changes

- Add amount formatter to handle big ints.
- Add create rate for token.
- Add exchange functionality with token.
- Support total supply.
- Able to view minted history

# Usage

- Go to token page, click view history to view minted history.
- Go to token detail and create rate.
- Transfer with different tokens.

# Impact

Deploy as usual.
screen shot 2018-07-05 at 4 44 32 pm


apps/admin_panel/assets/src/omg-alert/reducer.js
```diff
@@ -25,8 +25,8 @@ export const alertsReducer = createReducer([], {
   'ACCESS_KEY/CREATE/SUCCESS': state => {
     return [...state, createAlertState('Access key has successfully created.', 'success')]
   },
-  'ACCESS_KEY/CREATE/SUCCESS': state => {
-    return [...state, createAlertState('Access key has successfully created.')]
+  'EXCHANGE_PAIR/CREATE/SUCCESS': state => {
+    return [...state, createAlertState('Exchange pair has successfully created.', 'success')]
```
T-Dnzt: Exchange pair was successfully created.

apps/admin_panel/assets/src/omg-consumption-panel/index.js
```diff
@@ -8,13 +8,13 @@ import queryString from 'query-string'
 import { connect } from 'react-redux'
 import { approveConsumptionById, rejectConsumptionById } from '../omg-consumption/action'
 import { compose } from 'recompose'
-import { formatNumber } from '../utils/formatter'
+import { formatRecieveAmountToTotal } from '../utils/formatter'
```
T-Dnzt: recieve -> receive

apps/admin_panel/assets/src/omg-consumption-panel/index.js
```diff
@@ -93,6 +93,7 @@ class TransactionRequestPanel extends Component {
         consumptionId={queryString.parse(this.props.location.search)['show-consumption-tab']}
         render={({ consumption }) => {
           const tq = consumption.transaction_request || {}
+          console.log(consumption)
```
T-Dnzt: console.log

apps/admin_panel/assets/src/omg-create-transaction-modal/index.js
```diff
@@ -8,7 +8,7 @@ import { getWalletById } from '../omg-wallet/action'
 import { connect } from 'react-redux'
 import { compose } from 'recompose'
 import { withRouter } from 'react-router-dom'
-import { formatNumber } from '../utils/formatter'
+import { formatRecieveAmountToTotal, formatAmount } from '../utils/formatter'
```
T-Dnzt: Everywhere :')

T-Dnzt:
- Why can we set two rates for an exchange pair? It's only one, between token 1 and token 2
- Why did you call it rate? This is confusing since it may imply that you can create multiple rate for the same two tokens. Call it Exchange Pair so it matches what we have in the backend, we'll be adding helper tooltips to explain things if needed.
- I think we can remove the name from the rate form since we'll be removing it (to confirm with @unnawut)

PR #301, opened 2018-07-05
Controller authorization

T-Dnzt:
Issue/Task Number: T333

# Overview

This PR adds permissions for all controllers. Tests are very limited for now, but should be added. A bunch of endpoints should also be added to facilitate the retrieval of data for specific accounts (for example, getting `/transaction.all` will return all the transaction the current user/key has access to, this would need to be filtered for the current account in the admin panel either by passing a `search_terms` or by adding an endpoint: `/account.get_transactions`.

# Changes

- Add a policy for each controller
- Refactor code to support those policies
- Update `all` endpoints to return all the accessible records for the current user/key.

# Implementation Details

The code isn't super clean as you will see, I'm counting on the review to improve some of it, but will be spending some time next week to refactor and clean up. 

In many cases, I've used the approach of trying to load the record and returning `unauthorized` before running the permissions on it - I didn't want the policies to be responsible for loading those records if there was some action to be taken on them.

# Impact

This will probably impact the admin panel and clients, but shouldn't break anything. Let's do a round of testing to ensure that.

Tested with Robot (e2e repo) already, all passing.


apps/admin_api/test/admin_api/v1/controllers/admin_auth/transaction_controller_test.exs
```diff
@@ -157,64 +199,40 @@ defmodule AdminAPI.V1.AdminAuth.TransactionControllerTest do
...
+          "provider_user_id" => meta.user.provider_user_id
         })
 
-      assert response["data"]["object"] == "error"
-      assert response["data"]["code"] == "user:user_wallet_mismatch"
+      assert response["data"]["data"] |> length() == 8
 
-      assert response["data"]["description"] ==
-               "The provided wallet does not belong to the current user"
+      Enum.map(response["data"]["data"], fn tx ->
```
unnawut: Maybe `Enum.each/2` instead?

apps/admin_api/test/admin_api/v1/controllers/admin_auth/transaction_controller_test.exs
```diff
@@ -223,22 +241,14 @@ defmodule AdminAPI.V1.AdminAuth.TransactionControllerTest do
...
 
-      assert Enum.map(response["data"]["data"], fn t ->
-               t["id"]
-             end) == [
-               meta.transaction_4.id,
-               meta.transaction_3.id,
-               meta.transaction_2.id,
-               meta.transaction_1.id
-             ]
+      Enum.map(response["data"]["data"], fn tx ->
```
unnawut: Maybe `Enum.each/2` instead?

apps/admin_api/test/admin_api/v1/controllers/admin_auth/transaction_controller_test.exs
```diff
@@ -248,13 +258,11 @@ defmodule AdminAPI.V1.AdminAuth.TransactionControllerTest do
...
 
-      assert response["data"]["data"] |> length() == 1
+      assert response["data"]["data"] |> length() == 2
 
-      assert Enum.map(response["data"]["data"], fn t ->
-               t["id"]
-             end) == [
-               meta.transaction_4.id
-             ]
+      Enum.map(response["data"]["data"], fn tx ->
```
unnawut: Maybe `Enum.each/2` instead?

apps/admin_api/test/admin_api/v1/controllers/admin_auth/transaction_request_controller_test.exs
```diff
@@ -299,16 +307,8 @@ defmodule AdminAPI.V1.AdminAuth.TransactionRequestControllerTest do
...
-               "success" => false,
-               "version" => "1",
-               "data" => %{
-                 "code" => "wallet:wallet_not_found",
-                 "description" => "There is no wallet corresponding to the provided address",
-                 "messages" => nil,
-                 "object" => "error"
-               }
-             }
+      assert response["success"] == false
```
unnawut: Yes! Let's start asserting specific fields in the controller, rather than the full response map (since they're already covered by serializer tests and we are starting e2e tests now)

apps/ewallet_db/lib/ewallet_db/helpers/intersect.ex
```diff
@@ -0,0 +1,7 @@
+defmodule EWalletDB.Intersecter do
```
unnawut: :mindblown:

apps/ewallet_db/lib/ewallet_db/transaction_consumption.ex
```diff
@@ -299,9 +299,16 @@ defmodule EWalletDB.TransactionConsumption do
   end
 
   @spec query_all_for(Atom.t() | String.t(), any()) :: Ecto.Query.t()
+  def query_all_for(field_name, value) when is_list(value),
+    do: where(TransactionConsumption, [t], field(t, ^field_name) in ^value)
```
unnawut: `do ... end` for multilines?

apps/ewallet_db/test/ewallet_db/account_user_test.exs
```diff
@@ -0,0 +1,103 @@
...
+    test_has_valid_factory(AccountUser)
+  end
+
+  describe "AccountUser.insert/1" do
+    test_insert_generate_uuid(AccountUser, :uuid)
+    test_insert_generate_timestamps(AccountUser)
+    test_insert_prevent_blank(AccountUser, :account_uuid)
+    test_insert_prevent_blank(AccountUser, :user_uuid)
+
+    test "prevents inserting an account without an account" do
```
unnawut: "an account_user without an account"?

PR #302, opened 2018-07-05
Implement /transaction.calculate

unnawut:
Issue/Task Number: T486

# Overview

`/transaction.calculate` allows one to calculate the exchange transaction amounts without making an actual transaction.

This is useful for calculating the amounts to show to the user before making an actual transaction.

# Changes

- Added a new `TransactionCalculationController` and along with related view and serializer.
- Updated Open API specs

# Implementation Details

This new endpoint piggyback on `EWallet.Exchange` by managing different request parameters before sending into the `Exchange` module, then it receives an `EWallet.Exchange.Calculation` out and serializes it into a response.

# Usage

Try `/api/admin/transaction.calculate` by passing the following:
- `from_amount`
- `from_token_id` (required)
- `to_amount`
- `to_token_id` (required)

The request accepts a `from_amount` or `to_amount` or both:
- If only `from_amount` is provided, the `to_amount` will be calculated and returned using an existing exchange pair
- If only `to_amount` is provided, the `from_amount` will be calculated and returned using an existing exchange pair
- If both `from_amount` and `to_amount` are provided, the amounts will be compared to the existing exchange pair. If the amounts match with the exchange pair's rate, a successful calculation is returned. Otherwise it returns `exchange:invalid_rate` error.

# Impact

Code changes only. New API endpoint is added. No existing endpoints were changed.


apps/admin_api/lib/admin_api/v1/controllers/tranasction_calculation_controller.ex
```diff
@@ -0,0 +1,57 @@
...
+  @doc """
+  Calculates transaction amounts.
+  """
+  def calculate(conn, attrs) do
+    from_amount = attrs["from_amount"]
+    from_token = Token.get(attrs["from_token_id"])
+    to_amount = attrs["to_amount"]
+    to_token = Token.get(attrs["to_token_id"])
+
+    case do_calculate(from_amount, from_token, to_amount, to_token) do
```
T-Dnzt: You can pass `attrs["from_amount"]` and `attrs["to_amount"] directly I think.

apps/admin_api/lib/admin_api/v1/controllers/tranasction_calculation_controller.ex
```diff
@@ -0,0 +1,57 @@
...
+
+  defp do_calculate(nil, from_token, to_amount, to_token) do
+    Exchange.calculate(nil, from_token, to_amount, to_token)
+  end
+
+  defp do_calculate(from_amount, from_token, nil, to_token) do
+    Exchange.calculate(from_amount, from_token, nil, to_token)
+  end
+
+  defp do_calculate(from_amount, from_token, to_amount, to_token) do
```
T-Dnzt: Can't the two functions above be removed? Since they are just passing the same value for the same argument.

apps/admin_api/lib/admin_api/v1/controllers/tranasction_calculation_controller.ex
unnawut: Ah yes, i tried to cram all 4 params in but that looked very cluttered. Your suggestion looks like a good compromise. Updated.

apps/admin_api/lib/admin_api/v1/controllers/tranasction_calculation_controller.ex
unnawut: The two functions above go to `Exchange.calculate/4`, hence either allowed a nil from_amount or to_amount but not both. While the last one goes to `Exchange.validate/4`.

Could've done those two in one function with guards but it will look like this which looks confusing to me:

```ex
  defp do_calculate(from_amount, from_token, to_amount, to_token)
       when is_nil(from_amount) or is_nil(to_amount) do
    Exchange.calculate(nil, from_token, to_amount, to_token)
  end
```

I'll leave this one as is for this PR but open to more suggestions.

PR #311, opened 2018-07-06
Notify requester when consumption is finalized

T-Dnzt:
Issue/Task Number: A
 
# Overview

This PR adds the requester's channels to the list of channels notified when a consumption is finalized.


apps/ewallet/test/ewallet/schedulers/transaction_consumption_scheduler_test.exs
```diff
@@ -44,7 +44,7 @@ defmodule EWallet.TransactionConsumptionSchedulerTest do
       events = TestEndpoint.get_events()
 
       # 2 expired consumptions times 5 channels = 10 events
-      assert length(events) == 10
+      assert length(events) == 16
```
unnawut: Can you update the comment above it?

PR #312, opened 2018-07-06
Switch to Distillery in releases

### LINKED ISSUE #176: Dockerhub images: missing Hex and Rebar
# Dockerhub 

There seems to be a typo on Dockerhub which causes a syntax error.
`docker exec <container-id> mix ecto.create, ecto.migrate` should be `docker exec <container-id> mix do ecto.create, ecto.migrate`.

Also the current latest image doesn't seem to properly install `mix local.hex` and `mix local.rebar`. Since I get the following error:
```
docker exec d5967cfb5a10 mix do ecto.create, ecto.migrate
Could not find Hex, which is needed to build dependency :credo
Shall I install Hex? (if running non-interactively, use "mix local.hex --force") [Yn] ** (Mix) Could not find an SCM for dependency :credo from EWallet.Umbrella.Mixfile
```

# Environment

* OS: mac osx 10.12
* Docker: 18.03
* Docker-compose: 1.20

# Steps to reproduce
I used the following docker-compose.yml with `docker-compose up` and after that ran `docker exec <container-id> mix do ecto.create, ecto.migrate` in a seperate window.
```
version: '3.1'
services:
  db:
    image: postgres
  omisego:
    image: omisego/ewallet:latest
    ports:
      - 4000:4000
    environment:
      DATABASE_URL: "postgresql://postgres@db:5432/ewallet"
      LOCAL_LEDGER_DATABASE_URL: "postgresql://postgres@db:5432/local_ledger"
    links:
      - db
```

# Expected behavior
I expected the Dockerfile in omisego/ewallet:latest to properly install local.hex and local.rebar.

# Actual behavior
`$ docker exec d5967cfb5a10 mix do ecto.create, ecto.migrate`
```
Could not find Hex, which is needed to build dependency :credo
Shall I install Hex? (if running non-interactively, use "mix local.hex --force") [Yn] ** (Mix) Could not find an SCM for dependency :credo from EWallet.Umbrella.Mixfile
```

### LINKED ISSUE #249: Include an official docker-compose for the project
To simplify setup.

sirn:
* Fixes #176
* Fixes #249


sirn:
@T-Dnzt @mederic-p @unnawut This branch is now ready. Merry X’mas.

unnawut:
@sirn Does this address #598 too?

apps/ewallet/lib/ewallet/release_tasks.ex
```diff
@@ -0,0 +1,137 @@
...
+  defp priv_dir(app), do: "#{:code.priv_dir(app)}"
+
+  defp priv_path_for(repo, filename) do
+    app = Keyword.get(repo.config, :otp_app)
+    repo_underscore = repo |> Module.split() |> List.last() |> Macro.underscore()
+    Path.join([priv_dir(app), repo_underscore, filename])
+  end
+
+  #
+  # Config
```
unnawut: Can you explain here a bit when an app needs to be added to these lists? Both `@config_start_apps` and `@config_apps`.

Noticed that these apps are not part of the existing EWalletConfig's mix `extra_applications` nor in `EWalletConfig.Application`.

Dockerfile
```diff
@@ -1,79 +1,61 @@
...
-    echo '  -S mix omg.server --no-watch' >> $SERVICE_DIR/run && \
-    echo '#!/bin/execlineb -S1' > $SERVICE_DIR/finish && \
-    echo 'if { s6-test ${1} -ne 0 }' >> $SERVICE_DIR/finish && \
-    echo 'if { s6-test ${1} -ne 256 }' >> $SERVICE_DIR/finish && \
-    echo 's6-svscanctl -t /var/run/s6/services' >> $SERVICE_DIR/finish
-
-ENTRYPOINT ["/init"]
\ No newline at end of file
+EXPOSE $PORT
+EXPOSE 4369 6900 6901 6902 6903 6904 6905 6906 6907 6908 6909
```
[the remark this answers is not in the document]
sirn: They’re all for clustering, I think we should document about it in docs rather than a comment though. I’ll work on updating the docs next.

apps/ewallet/lib/ewallet/release_tasks.ex
sirn: I'm not quite sure what would be the best way to document these apps. They're not in `extra_applications` due to how Mix/Elixir inferred extra apps based on what's defined in `defp deps`. In case of running release tasks, since we'll be invoking the code directly, we need to explicitly define them since there's no Mix deps to defer from. Try running `make build-prod` and inspect the generated `ewallet.app` file it generated, it should contains the full extra applications that Distillery deferred from Mix during compilation.

PR #328, opened 2018-07-12
Generate JSON specs on the fly

unnawut:
Issue/Task Number: T430

# Overview

This PR enables the OpenAPI json specs to be generated automatically when the YAML specs change.

# Changes

- Add `EWallet.Web.APIDocs.JSONGenerator`
- Add the generator to `AdminAPI.Application` and `EWalletAPI.Application` so the yaml specs are monitored

# Implementation Details

- This feature adds a new dependency: `:file_system`, which is also used by [`:phoenix_live_reload`](https://github.com/phoenixframework/phoenix_live_reload/blob/master/mix.exs#L42)
- The generator is implemented as a `GenServer` so that it does not block the ewallet server

# Usage

Run `mix omg.server` and try touch or update the spec.yaml files. The server should output the following:

```ex
$ mix omg.server
[info] Setting up websockets dispatchers...
[info] Running UrlDispatcher.Plug with Cowboy http on port 4000
[info] Changes detected at /vagrant/apps/ewallet_api/priv/spec.yaml. Regenerating JSON API specs...
[info] JSON OpenAPI spec generated successfully at /vagrant/apps/ewallet_api/priv/spec.json
```

# Impact

No DB changes. No API changes.


sirn:
Not sure if I like the idea of adding Java as runtime dependencies (= we will need it everywhere), especially when the spec is pretty much static? 🤔 I think having a [pre-commit hook](https://git-scm.com/book/gr/v2/Customizing-Git-Git-Hooks) and force anyone who modifies the YAML spec to generate JSON spec (and commit that) might be a better approach.

unnawut:
Imho, I don't trust (myself with) pre-commit hooks as it relies on the committer setting up the hook and not skip it. Personally it would be nice if we could at least generate it during the build and not runtime?

Or maybe it's time to look for alternative to OpenAPI together given the limitations we're discovering... 🤔

PR #336, opened 2018-07-16
Add user test

jarindr:
Issue/Task Number: 508-refactoring-everything-and-add-tests

# Overview
Add test for users and current user.

# Changes
- Add test for users and current user action.
- Add test for users and current reducer.
- Renaming.
- Fix minor typo bugs.

# Usage
run `yarn test`

# Impact
deploy as usual

### DIFF
--- a/apps/admin_panel/assets/src/omg-user-current/action.js
+++ b/apps/admin_panel/assets/src/omg-user-current/action.js
@@ -1,11 +1,11 @@
-import * as currentUserSerivce from '../services/currentUserService'
+import * as currentUserService from '../services/currentUserService'
 import * as adminService from '../services/adminService'
 import { createActionCreator } from '../utils/createActionCreator'
 export const getCurrentUser = () =>
   createActionCreator({
     actionName: 'CURRENT_USER',
     action: 'REQUEST',
-    service: currentUserSerivce.getCurrentUser
+    service: currentUserService.getCurrentUser
   })
 
 export const updateCurrentUser = ({ avatar, email }) =>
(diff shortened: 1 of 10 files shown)


apps/admin_panel/assets/src/omg-user-current/currentUserProvider.js
```diff
@@ -17,7 +17,7 @@ class UserProvider extends Component {
   componentDidMount = async () => {
     if (this.props.currentUserLoadingStatus === 'DEFAULT') {
       const result = await this.props.getCurrentUser()
-      if (!result.data.success) {
+      if (!result.data) {
```
T-Dnzt: Why the change? Shouldn't you be checking the error code as well? Or maybe it's displayed after the push?

apps/admin_panel/assets/src/omg-user-current/action.js
```diff
@@ -1,51 +1,27 @@
...
-  } catch (error) {
-    return dispatch({ type: 'CURRENT_USER/REQUEST/FAILED', error })
-  }
-}
+import { createActionCreator } from '../utils/createActionCreator'
+export const getCurrentUser = () =>
+  createActionCreator({
+    actionName: 'CURRENT_USER',
+    action: 'REQUEST',
+    service: currentUserSerivce.getCurrentUser
```
T-Dnzt: service, not serivce :)

apps/admin_panel/assets/src/omg-user-current/currentUserProvider.js
jarindr: This is a data provider component, we don't need to handle any error code here.
if the returned result has data, it is guaranteed to be a success action. if it is error, it will return `result.error`

Issue #341, opened 2018-07-18
Switch to use DeferredConfig to manage environment variables


sirn:
Use [deferred_config](https://github.com/mrluc/deferred_config) for compatibility with Distillery where config.exs will be evaluated at build time.

sirn:
I'm going to extract this from #312 since working on new building pipeline is taking wayyyy too long and this was done for months and we haven't tested it. 🤦‍♂️

PR #343, opened 2018-07-18
Switch to use DeferredConfig

sirn:
Closes: #341 

# Overview

Switch to use DeferredConfig for configurations so that it works with Distillery which evaluate config at build time rather than during runtime.  This means anything that involve `System.get_env` cannot be used.  DeferredConfig solve this by using a configuration tuple `{:system, NAME}` and rewrite it to read from system env during application launch.

This branch is extracted from #312 

# Impact

Needs extensive testing.


apps/ewallet/lib/ewallet/web/paginator.ex
```diff
@@ -67,7 +67,20 @@ defmodule EWallet.Web.Paginator do
   # Returns the per_page number or default, but never greater than the system's defined limit
   defp get_per_page(attrs) do
     per_page = Map.get(attrs, "per_page", @default_per_page)
-    max_per_page = Application.get_env(:ewallet, :max_per_page, @default_max_per_page)
+
+    max_per_page =
+      case Application.get_env(:ewallet, :max_per_page, @default_max_per_page) do
```
unnawut: How come this one we need to handle the `{:system, _}` tuple ourselves while other configs don't need to?

apps/ewallet/lib/ewallet/release_tasks.ex
```diff
@@ -0,0 +1,88 @@
+defmodule EWallet.ReleaseTasks do
```
unnawut: Just making sure... this is part of this PR? Looks like it's a part of #312 or even its own PR?

sirn: Nope, I forgot to delete it :\

sirn:
> So, any edits to the env vars after the application has started will have no effects right? So applications should be restarted after editing the env vars to take effect.

This was always the case; you cannot change configuration and have it apply during runtime. 

> Also I noticed some uses of `System.get_env/1` still. We should continue to use `System.get_env/1` for runtime code and DeferredConfig for compile-time code (config/*.exs)?

Yup. All `config/*.exs` should use DeferredConfig. The rest (in the app itself) can use `System.get_env` as always.

apps/ewallet/lib/ewallet/web/paginator.ex
sirn: Because `ewallet` the module isn't an app so we don't have exact timing to call `DeferredConfig`.

sirn:
@T-Dnzt any comments?

PR #344, opened 2018-07-18
Add enable disable functionality for access key / Refine design

jarindr:
Issue/Task Number: - 

# Overview

This PR add enable disable functionality for access key and refine some design.

# Changes

- Add enable disable functionality for access key.
- Refine some design.

# Impact

Deploy as usual.


apps/admin_panel/assets/src/omg-page-api/index.js
```diff
@@ -179,8 +187,34 @@ class ApiKeyPage extends Component {
         />
       )
     }
-    if (key === 'status_access') {
-      return data ? 'disabled' : 'enabled'
+    if (key === 'key') {
```
mederic-p: can't we use a switch/case here instead?

Issue #373, opened 2018-07-24
Transaction Request Exchange are not handled properly in Admin Panel


T-Dnzt:
When specifying two different tokens (creating a request on device with Token 1 and consuming it in admin panel with Token 2), it will fail saying the `exchange_account_id` or `exchange_wallet_address` is missing.

PR #398, opened 2018-08-01
Enable transaction request exchange functionality.

jarindr:
Issue/Task Number: #373 

# Overview

Since transaction request creation can't input exchange address before, it is not possible to consume with different token. This PR add a field when creating transaction request.

# Changes

- Add exchange address transaction request.
- Change loading style for table.
- Change loading style for button.
- Add exchange address in detail.
- Refactor components to share transaction request detail component.
- Add calculate action and use for calculating consume different token.
- Fix consumption amount is wrong when use with exchange pair.
- Minor bug fix.

# Usage

- Move around to see new loading style.
- Create transaction request with exchange address, consume it with different token.
- Save something to see loading button style.

# Impact

Deploy as usual.


mederic-p:
screen shot 2018-08-06 at 09 55 39
Amount is incorrect when consuming a different token. (wrong token and wrong subunit to unit)

I created a new account but it didn't show up in the list of addresses that I can pick for the `Exchange address`. I had to reload the page.

When consuming a request that DOESN'T allow amount override: it'd be nice to update the amount box when selecting different token using the `calculate_amount` endpoint

Also would be nice to have the exchange address in the `additional request details`

Also I think the consumption box should be redesigned to have a better view of the exchange (to discuss with Noei)

apps/admin_panel/assets/src/omg-create-transaction-request-modal/index.js
```diff
@@ -123,7 +135,8 @@ class CreateTransactionRequest extends Component {
         })
       }
     } catch (e) {
-      this.setState({ submitting: false, error: e })
+      console.log(e)
```
T-Dnzt: console.log

Issue #369, opened 2018-07-24
Sync option for exchange pairs is missing


T-Dnzt:
Exchange pairs can be synced when creating/updating. This is currently missing from the exchange pair form.

PR #400, opened 2018-08-02
Add syncing exchange pair

jarindr:
Issue/Task Number: #369 

# Overview

Add sync option for exchange pair

# Changes

- Add sync option for exchange pair

# Usage

Go to token, create an exchange pair and tick sync, now the rate should apply to both way.

# Impact

Deploy as usual.


jarindr:
@mederic-p actual rate needs service from #398 I will add another PR for it

PR #401, opened 2018-08-03
Allow the eWallet to work without integration

### LINKED ISSUE #357: Allow the eWallet to work without integration
Currently, the eWallet cannot be used as a stand-alone with only mobile applications using it (currently, it is needed for a provider to create an application to store users, etc.)

This feature is needed for the sample Coffee shop application we're going to be working on.

It involves:

- A merchant app, where an administrator can login (by entering email/password). From this application, a logged-in admin can then proceed to top-up or purchase on behalf of a user by scanning a user's identity QR code. This application should automatically logout admins unless they specified a "Remember Me" button, hence the need to add Face ID / Touch ID.
- A user app, where a user should be able to sign up and login (all through the eWallet API, `/signup` is currently missing. The user can then see his balances or generate his identity QR code.

Tasks:

- [ ] Ensure the differentiation field between admins and users is only `provider_user_id`
-  [ ] Add `/user.signup` endpoint in client API, which will only proceed a request if `ALLOW_USER_REGISTRATION` is set to `true` (defaults to `false`)
- [ ] Add `/user.login` endpoint in client API (which follows the same rules as above, depending on `ALLOW_USER_REGISTRATION`

unnawut:
Closes #357.

# Overview

This PR allows an eWallet provider to setup and uses the eWallet without having to implement or integrate into their own solutions.

# Changes

## Email configs
- Moved Bamboo and email-related configs from `AdminAPI` to `EWallet` app (to share with `EWalletAPI`).
- Generalized `AdminAPI.Inviter` into `EWallet.Web.Inviter` so that it accepts the email template as a parameter. So AdminAPI and EWalletAPI can send the confirmation email with different template/messages.

## Authentication
- Updated `AdminAPI.V1.AdminUserAuthenticator` to reject authentication of non-admin users
- Added `EWalletAPI.V1.EndUserAuthenticator` to authenticate a user with email and password

## Error handling
- Moved `:invalid_login_credentials`, `:invalid_email`, `:invite_not_found`, `:passwords_mismatch` from `AdminAPI.V1.ErrorHandler` to `EWallet.V1.ErrorHandler`

## Signup flow
- Added `/api/client/user.signup` to support user signup with an email through the new `EWalletAPI.V1.SignupController.signup/2`
- Added `/api/client.user.verify_email` to support user email verification through the new `EWalletAPI.V1.SignupController.verify_email/2`
- Added `/api/client/user.login` to support user login with email/password through the new `EWalletAPI.V1.AuthController.login/2`

## Supporting signup flow
- Added a new `EWalletAPI.VerificationEmail` template, separated from `AdminAPI.InviteEmail`
- Updated `EWallet.Web.V1.UserAuthTokenSerializer.serialize/1` to return user along with the authentication token
- Added `:invite_uuid` to `User.update_changeset/2` so that invites can be removed after verification
- Updated `EWalletDB.User.has_membership?/1` to also handle user's UUID as the parameter.
- Added `EWalletDB.User.admin?/1`

## Docs
- Updated OpenAPI spec with `/user.login`, `/user.signup`, `/user.verify_email`
- Added `docs/design.users.md` to explain different user types and the checker functions
- Added typespecs for `EWalletDB.User` functions

# Implementation Details

Since this functionality can be built on the existing invitation system for admins (and admins are also users), most of the changes are for reusing the existing code.

# Usage

Try sign up with `/api/client/user.signup` followed by `/api/client.user/verify_email`, then `/api/client/user.login`.

## Example curl for /user.sign_up

```
curl 10.5.10.10:4000/api/client/user.signup \
-d '{
"email": "test_signup@example.com",
"redirect_url": "http://localhost:4000/api/client/user.verify_email?email={email}&token={token}"
}' \
-v -w "\n" | jq
```

## Example curl for /user.verify_email

```
curl 10.5.10.10:4000/api/client/user.verify_email \
-d '{
"email": "test_signup@example.com",
"token": "kXN6Kd6HaJiYabpg7LWM3FCCHBMMLRKMBnIkMxxqjE4",
"password": "the_password",
"password_confirmation": "the_password"
}' \
-v -w "\n" | jq
```

## Example curl for /user.login

```
curl 10.5.10.10:4000/api/client/user.login \
-d '{
"email": "test_signup@example.com",
"password": "the_password",
}' \
-v -w "\n" | jq
```

# Impact

- Run `mix ecto.migrate` to migrate the database
- Set the environment variable `ENABLE_STANDALONE=true` to enable standalone
- Optionally add `REDIRECT_URL_PREFIXES` as a comma-separated string to allow redirect URLs that are external to the eWallet. If a redirect URL is already prefixed with `BASE_URL`, it does not need to be added here.


T-Dnzt:
@unnawut Can you change this to allow password/password_confirmation to be sent in in `/user.sign_up` as optional parameters? And send back an error if they are sent again in the `verify` step.

unnawut:
> And send back an error if they are sent again in the verify step.

If the user tries to verify their email and tries to set the passwords again, he/she will receive an error in this case, and they will be stuck. I propose that the passwords set in `/user.verify_email` should overwrite the existing password (i.e. the one set at `/user.sign_up`).

T-Dnzt:
@unnawut Okay with the override. Or just remove the ability to pass the password in the `verify_email`, after thinking about it, the only use-case would be users inviting other users. Do we want that? If yes, let's keep it, if not we can remove it.

Other stuff to add:

- [ ] ENV variable to toggle on/off those two endpoints (will later be moved in `Setting`)
- [ ] Add a default blank page with one sentence saying `Your account is confirmed.`. This should be the default when no `redirect_url` is provided in the `sign_up` step. This can be accessible under `/pages/user_confirmation` maybe?
- [ ] Add a new parameter `confirmed_redirect_url`, that will be called automatically after `verify_email` if all is good in the world (password is set, user is confirmed). If not specified, the user should be taken to the page mentioned above.

unnawut:
Another task:
- [ ] A successful signup (which are unauthenticated) should not return the user data as we should not reveal those info to unauthenticated users.

unnawut:
To check:
- Code that requires `%{"provider_user_id" => conn.assigns.user.provider_user_id}` which won't be available from standalone users
- Attaching a user to an account on user creation
- Check that user has a wallet created

unnawut:
> Or just remove the ability to pass the password in the verify_email, after thinking about it, the only use-case would be users inviting other users. Do we want that? If yes, let's keep it, if not we can remove it.

Just realized that setting password in `verify_email` step will complicate the flow a bit. E.g. we need to ask the user to set a password before reaching the default "Your account is confirmed." page. 

So I'll make `password` and `password_confirmation` mandatory during `/user.signup` and remove them from `/user.verify_email`.

T-Dnzt:
@unnawut A few issues I found this morning while testing. Please discard if they have already been fixed.

- [ ] Existing admin users can't login anymore due to `is_admin` not being set to `true`
- [ ] Seeded administrators don't have `is_admin` set to true
- [ ] Can't signup with a user, getting `endpoint_not_found` from Swagger

mederic-p:
Also, for now if a registration fails because of an invalid url, we won't be able to submit it again because the user is already created in the database.
I can see that you validate the url in 2 different files: `signup_gate.ex` and `inviter.ex` but in 2 different way. You may want to refactor this.

unnawut:
Remaining:
- [ ] validate `success_url` before creating the user
- [ ] Swagger update
- [ ] Deprecate `redirect_url` in favor of `validation_url`
- [ ] `validation_url` & `success_url` whitelist that supports http(s) + deep linking
- [ ] Refactor `validation_url` and `success_url` validation to the same level
- [ ] Refine error to "Your user account hasn't been confirmed yet. Please check your emails."
- [ ] Retest all endpoints

apps/admin_api/lib/admin_api/v1/auth/admin_user_authenticator.ex
```diff
@@ -4,12 +4,22 @@ defmodule AdminAPI.V1.AdminUserAuthenticator do
...
-    authenticate(conn, user, password)
+    with %User{} = user <- User.get_by_email(email) || :user_email_not_found,
+         true <- User.admin?(user) || :user_not_admin do
+      authenticate(conn, user, password)
+    else
+      :user_email_not_found ->
+        Crypto.fake_verify()
+        handle_fail_auth(conn, :invalid_login_credentials)
+
+      :user_not_admin ->
```
T-Dnzt: Any difference between this case and the one above? Can't we juse use:

```
_ -> 
  Crypto.fake_verify()
  handle_fail_auth(conn, :invalid_login_credentials)
```

apps/admin_api/lib/admin_api/v1/controllers/admin_auth_controller.ex
```diff
@@ -3,53 +3,43 @@ defmodule AdminAPI.V1.AdminAuthController do
...
+           AuthToken.get_by_token(token, :admin_api) || {:error, :auth_token_not_found},
          {:ok, token} <- AuthToken.switch_account(token, account) do
       render_token(conn, token)
     else
-      {:error, error} ->
-        handle_error(conn, error)
-
-      error ->
-        handle_error(conn, error)
+      {:error, code} when is_atom(code) ->
```
T-Dnzt: 💯

apps/admin_api/test/admin_api/v1/controllers/admin_auth/account_membership_controller_test.exs
```diff
@@ -257,7 +257,7 @@ defmodule AdminAPI.V1.AdminAuth.AccountMembershipControllerTest do
       assert response["data"]["code"] == "client:invalid_parameter"
 
       assert response["data"]["description"] ==
-               "The `redirect_url` is not allowed to be used. Got: '#{redirect_url}'."
+               "The given `redirect_url` is not allowed to be used. Got: '#{redirect_url}'."
```
T-Dnzt: Can we change this error to -> "The given `redirect_url` is not allowed. Got: '#{redirect_url}'." ?

apps/ewallet/lib/ewallet/application.ex
```diff
@@ -7,6 +7,7 @@ defmodule EWallet.Application do
 
   def start(_type, _args) do
     import Supervisor.Spec
+    DeferredConfig.populate(:ewallet)
```
T-Dnzt: Nice!

apps/ewallet_api/lib/ewallet_api/emails/verification_email.ex
```diff
@@ -0,0 +1,39 @@
...
+    |> from(sender)
+    |> subject("OmiseGO eWallet: Verify your email")
+    |> html_body(html(link))
+    |> text_body(text(link))
+  end
+
+  defp html(link) do
+    """
+    <p>
+      <strong>Click the link to complete the email verification: </strong>
```
T-Dnzt: `to complete the email verification process`

apps/ewallet_api/lib/ewallet_api/emails/verification_email.ex
```diff
@@ -0,0 +1,39 @@
...
+      <a href="#{link}">#{link}</a>
+    </p>
+    """
+  end
+
+  defp text(link) do
+    """
+    Copy & paste the link into your browser to complete the email verification: #{link}
```
T-Dnzt: `email verification process`

apps/ewallet_api/lib/ewallet_api/v1/controllers/pages/verify_email_controller.ex
```diff
@@ -0,0 +1,44 @@
...
+      {:ok, %{success_url: nil}} ->
+        redirect(conn, to: verify_email_path(conn, :success))
+
+      {:ok, %{success_url: success_url}} ->
+        redirect(conn, external: success_url)
+
+      {:error, error} ->
+        case Map.get(ErrorHandler.errors(), error) do
+          %{description: description} ->
+            text(conn, "Unable to verify your email address. " <> description)
```
T-Dnzt: `We were unable to verify your email address. Error description: `.

apps/ewallet_api/lib/ewallet_api/v1/controllers/pages/verify_email_controller.ex
```diff
@@ -0,0 +1,44 @@
...
+
+          nil ->
+            text(conn, "Unable to verify your email address. An unknown error occured.")
```
T-Dnzt: `We were unable to verify your email address.`

apps/ewallet_api/lib/ewallet_api/v1/controllers/pages/verify_email_controller.ex
```diff
@@ -0,0 +1,44 @@
...
+        end
+    end
+  end
+
+  @doc """
+  Renders the page to show when email verification is successful.
+  """
+  def success(conn, _attrs) do
+    text(conn, "Your email address is successfully verified!")
```
T-Dnzt: `Your email address has been successfully verified!`

apps/ewallet_db/lib/ewallet_db/account.ex
```diff
@@ -259,14 +259,30 @@ defmodule EWalletDB.Account do
   @doc """
   Get the master account for the current wallet setup.
   """
-  @spec get_preloaded_primary_wallet(keyword()) :: %Account{} | nil
+  @spec get_master_account(keyword()) :: %Account{} | nil
```
T-Dnzt: 🎉

apps/ewallet_db/lib/ewallet_db/invite.ex
```diff
@@ -93,6 +115,22 @@ defmodule EWalletDB.Invite do
...
+  @doc """
+  Accepts an invitation without setting a new password.
+  """
+  @spec accept(%Invite{}) :: {:ok, struct()} | {:error, any()}
+  def accept(invite) do
+    invite = Repo.preload(invite, :user)
+
+    case User.update_without_password(invite.user, %{invite_uuid: nil}) do
+      {:ok, _user} ->
+        Repo.delete(invite)
```
T-Dnzt: Do we actually delete invites or is it soft-delete? Can't we just set them to `used` or something and keep them?

apps/admin_api/lib/admin_api/v1/auth/admin_user_authenticator.ex
unnawut: You're right. Midway it was a bit different but not anymore. Updating.

apps/ewallet_db/lib/ewallet_db/invite.ex
unnawut: It’s hard delete right now. To do soft delete, I'll do this:

1. Disassociate the invite from the user (`user.invite_uuid`) after verification. So `user.invite_uuid == nil`.
2. Add a new `invite.user_uuid` column to keep the user->invite trace after the disassociation from `user.invite_uuid`.
3. Add a new `invite.verified_at` column so it’s easy to see if and when the invite was used.

unnawut:
Updated with @T-Dnzt 's comments as well as @mederic-p 's suggestion to auto-include `BASE_URL` as a `redirect_url_prefixes`

PR #424, opened 2018-08-16
strict dialyzer analysis

InoMurko:
Issue/Task Number:

# Overview

Added a more strict Dialyzer setting.

# Changes

Fixed the reported issues from master.

# Implementation Details

Fixed specs 

# Usage

/

# Impact

Nothing mayor.


unnawut:
This should be able to go with v1.0 since it's only `@spec` changes

InoMurko:
if you added :specdiffs to the configuration things would look different :-) I suggest you try it out and decide whether you think it's useful.

unnawut:
Thank you for your PR! 🎖🥇🥈🍾 🌟

PR #440, opened 2018-08-28
Composable filters as request params

### LINKED ISSUE #334: Implement advanced filtering for Admin API
Currently, it isn't possible to do proper advanced filtering using `search_term` and `search_terms`. An advanced filtering system needs to be implemented (potentially using a new field, `filters`?) that would contain a list of `AND`/`OR` predicates.

The filtering system should allow association filtering up to one level (e.g. from `/wallet.all`, I should be able to specify `account.name = something`. Different comparison mechanism should be supported: `greater`, `equal`, etc.).

Open to suggestions for Elixir libraries that could handle this.

unnawut:
Closes #334 

# Overview

This PR adds support for a more advanced filtering. Instead of just `search_term` and `search_terms`, it's able to filter by conditions like `eq` (equals), `neq` (not equals), `lt` (less than), `gte` (greater than), etc.

# Changes

- Add a new `EWallet.Web.MatchAllParser` to support filtering with all conditions
- Add a new `EWallet.Web.MatchAnyParser` to support filtering with any conditions

# Implementation Details

This PR took inspirations from various existing open-sourced libraries but decided to write from scratch as the libraries are either very feature-rich that we use a very small subset of the feature, or very simple that it could easily be built on our existing parsers.

Also note that "match_any" is limited to 5 unique associations. This is due to the fact that Ecto supports only positional reference on joins. [Named joins](https://github.com/elixir-ecto/ecto/issues/2389) is coming in Ecto 3.0 but the release date is not set. So, I did a workaround by mapping the join sequence and build the dynamic query by having a fixed number of bindings (fixed to 5).

# Usage

Supported comparators: `eq`, `neq`, `lt`, `lte`, `gt`, `gte`, `contains` and `starts_with`.

For example, the attributes below will filter for confirmed transactions that came from or went to the user with username "alice".

```js
POST /transaction.all
{
  "match_all": [
    {
      "field": "status",
      "comparator": "eq",
      "value": "confirmed"
    }
  ],
  "match_any": [
    {
      "field": "from_user.username",
      "comparator": "eq",
      "value": "alice"
    },
    {
      "field": "to_user.username",
      "comparator": "eq",
      "value": "alice"
    }
  ]
}
```

# Impact

New optional attributes: `match_all` and `match_any` added to `/transaction.all`'s request format. No DB changes.


unnawut:
Remaining:
- [ ] Support field value type casting similar to `{field, :uuid}` similar to the existing `SearchParser`
- [ ] Support nested field (e.g. filter transactions by `from_account.name`, etc.)
- [ ] Support predicate e.g. `match: all` or `match: any`

unnawut:
Example for multiple predicates:

```
POST /transaction.all
{
  "filters": {
    "match_all": [
      {
        "field": "from_user.id",
        "comparator": "eq",
        "value": "alice_id"
      },
      {
        "field": "to_user.id",
        "comparator": "eq",
        "value": "bob_id"
      }
    ],
    "match_any": [
      {
        "field": "status",
        "comparator": "eq",
        "value": "confirmed"
      },
      {
        "field": "status",
        "comparator": "eq",
        "value": "approved"
      }
    ]
  }
}
```

T-Dnzt:
@unnawut Could you add more tests? Like at least one for each predicate, and maybe test up to 5 levels?

unnawut:
@T-Dnzt Added `starts_with`, OpenAPI specs and markdown docs. Please review again.

PR #443, opened 2018-08-30
Allows tokens & wallets to be disabled

### LINKED ISSUE #434: Allow tokens /wallets to be disabled.

T-Dnzt:
Issue/Task Number: 434
closes #434 

# Overview

Since we can't delete tokens and wallets, we need to have the next best thing: an `enabled` property! Once disabled (`enabled = false`), a token/wallet will still be shown, but will not be usable for anything, unless it is re-enabled later.

# Changes

- Add `enabled` to tokens
- Add `enabled` to wallets
- Add new endpoints: `/token.enable_or_disable` and `/wallet.enable_or_disable`
- Add tests at the controller and gate levels


apps/admin_api/test/admin_api/v1/controllers/admin_auth/wallet_controller_test.exs
```diff
@@ -591,4 +594,62 @@ defmodule AdminAPI.V1.AdminAuth.WalletControllerTest do
       assert length == 3
     end
   end
+
+  describe "/wallet.disable" do
```
unnawut: /wallet.enable_or_disable

unnawut:
Notes for future reference:

This enable/disable switches are under their own endpoints and not part of i.e. `/token.update` because they have a significant impact on the overall system (e.g. cannot perform transactions, etc.)

PR #452, opened 2018-09-12
Update Docker setup instructions

unnawut:
Issue/Task Number: -

# Overview

Update the docker-compose setup instructions following the feedback from the Shanghai workshop.

# Changes

- Wrapped the elixir command for secret key generation inside `docker run` (so we don't require elixir on the host machine in the docker approach)
- Added `-it` flag to mix commands as a workaround for being prompted for hex/rebar3 install (to be removed when it's fixed at the docker image level.

# Implementation Details

N/A

# Usage

N/A

# Impact

N/A


docs/setup/docker.md
```diff
@@ -65,16 +59,18 @@ volumes:
   postgres-db:
 ```
 
-Then, run the following command to provision the images:
+Notice that the values for `EWALLET_SECRET_KEY` and `LOCAL_LEDGER_SECRET_KEY` are missing. You can generate these secret keys by running:
 
-```bash
-$ docker-compose up
 ```
+$ docker run --rm omisego/ewallet:latest elixir -e "IO.puts 32 |> :crypto.strong_rand_bytes() |> Base.encode64()"
```
sirn: Might worth generating keys with `openssl rand -base64 32` instead so user don't need to have Elixir install, or have a running instance of eWallet (which is the recommended way to do so in [Helm chart](https://github.com/omisego/charts/blob/3033de464aaea4210af4d23fb1972c03aab239ed/ewallet/values.yaml#L58)).

T-Dnzt: That's a very good point, realised that it was kinda of annoying for people to have to download Elixir just for this.

unnawut:
@sirn @T-Dnzt Please review again.

Note that due to v1.0.0 pinning, I also replaced this text:

> While the command above pulls the latest image, *it is highly recommended to pin the version to specific commit in anything that resembles a production environment.*

to

> The command above pulls the v1.0.0 image. If you wish, you can pick [other available tags from the Docker Hub](https://hub.docker.com/r/omisego/ewallet/tags/).

PR #457, opened 2018-09-18
User tracking/auditing

### LINKED ISSUE #364: Logging/Tracking of who did what on what
In order to proceed with blockchain integration, we need to finally add this feature we've been pushing to "later". With this, we'll have a log of all actions that happened in the system and we will be able to find out who did what and when.

T-Dnzt:
Issue/Task Number: #364
Closes #364

# Overview

This PR adds the foundation for an audit system, starting with a focus on users.

# Changes

- Track user and invite creation/update
- Prevent `ForgetPasswordRequest` from being deleted - instead they should only be disabled.
- Add the `Originator` logic

# Implementation Details

When creating records, `Audit.insert` or `Audit.update` should be used instead, it will generate an `audit` that will be inserted in a SQL transaction. It is possible to pass an `Ecto.Multi` if more records need to be inserted (like wallets when creating users).


apps/ewallet_db/lib/ewallet_db/types/virtual_struct.ex
```diff
@@ -0,0 +1,23 @@
+defmodule EWalletDB.Types.VirtualStruct do
```
unnawut: Ummm.. Can you explain a bit more why we need VirtualStruct? Strings won't work? :S

apps/ewallet_db/lib/ewallet_db/user.ex
```diff
@@ -228,19 +236,19 @@ defmodule EWalletDB.User do
...
   @spec insert(map()) :: {:ok, %User{}} | {:error, Ecto.Changeset.t()}
   def insert(attrs) do
-    multi =
-      Multi.new()
-      |> Multi.insert(:user, changeset(%User{}, attrs))
-      |> Multi.run(:wallet, fn %{user: user} ->
-        case User.admin?(user) do
+    %User{}
+    |> changeset(attrs)
+    |> Audit.insert(
```
unnawut: Maybe a future refactor suggestion...

The naming `Audit.insert()`, `Audit.update()`, etc. makes me think that it's the audit trail insert only, and not the user.

Could we make an Ecto facade, e.g. `AuditedEcto`, and change from `alias Ecto.Multi` to `alias EWalletDB.AuditedEcto.Multi` and so we could keep the same `Multi.insert()`?

Probably easier said than done 😅

apps/ewallet_db/lib/ewallet_db/types/virtual_struct.ex
T-Dnzt: Well, it's to pass the actual struct without worrying about its type or having to add a formatter I guess. What would you suggest?

apps/ewallet_db/lib/ewallet_db/user.ex
T-Dnzt: Hmmm, I don't really want to override an Ecto module with our own, makes it quite confusing I think 🤔 Maybe I'm not getting the idea tho. But I can rename the function to make it clearer.

apps/ewallet_db/lib/ewallet_db/audit.ex
```diff
@@ -61,6 +65,10 @@ defmodule EWalletDB.Audit do
     Application.get_env(:ewallet_db, :schemas_to_audit_types)[schema]
   end
 
+  def all_for_target(record) do
+    Repo.all(Audit, target_type: get_type(record.__struct__), target_uuid: record.uuid)
```
unnawut: I just realized this. I think the 2nd and 3rd params have no effect: https://hexdocs.pm/ecto/Ecto.Repo.html#c:all/2

Same goes to the function below.

apps/ewallet_db/test/support/factory.ex
```diff
@@ -148,7 +148,7 @@ defmodule EWalletDB.Factory do
 
   def audit_factory do
     params = params_for(:user)
-    user = insert(params)
+    user = insert(:user, params)
```
unnawut: 😅

Issue #439, opened 2018-08-27
Cannot login to admin panel using docker-compose version


LyXman:
# Subject of the issue

I have setup the ewallet as described in the docs using the docker-compose file approach. Everything goes well and the sample database is created, but when I try to login with the self-made credentials following error occurs every time:

ewallet_1   | 16:35:58.494 request_id=2l7cm20hcoc05ldga80005n1 [info] GET /admin
ewallet_1   | 16:35:58.593 request_id=2l7cm20hcoc05ldga80005n1 [info] Sent 200 in 99ms
ewallet_1   | 16:36:08.673 request_id=2l7cm2jg7leusgo60g0005oh [info] POST /api/admin/admin.login
ewallet_1   | 16:36:08.796 request_id=2l7cm2jg7leusgo60g0005oh [info] Sent 500 in 123ms
ewallet_1   | 16:36:08.801 [error] #PID<0.1165.0> running UrlDispatcher.Plug terminated
ewallet_1   | Server: localhost:4000 (http)
ewallet_1   | Request: POST /api/admin/admin.login
ewallet_1   | ** (exit) an exception was raised:
ewallet_1   |     ** (ArgumentError) argument error
ewallet_1   |         :erlang.iolist_to_binary(:error)
ewallet_1   |         (poison) lib/poison/parser.ex:35: Poison.Parser.parse/2
ewallet_1   |         (poison) lib/poison/parser.ex:51: Poison.Parser.parse!/2
ewallet_1   |         (poison) lib/poison.ex:83: Poison.decode!/2
ewallet_1   |         (ewallet_db) lib/cloak/field.ex:48: EWalletDB.Encrypted.Map.load/1
ewallet_1   |         (ecto) lib/ecto/type.ex:647: Ecto.Type.process_loaders/3
ewallet_1   |         (ecto) lib/ecto/schema.ex:1416: Ecto.Schema.load!/5
ewallet_1   |         (ecto) lib/ecto/schema.ex:1399: Ecto.Schema.__load__/5

# Your environment

OS: Ubuntu 16.04 LTS
Docker: 18.03.1-ce, build 9ee9f40
Docker-Compose: 1.13.0, build 1719ceb

# Other log entries from the docker images

mailhog_1   | [APIv1] KEEPALIVE /api/v1/events
postgres_1  | FATAL:  role "root" does not exist

unnawut:
[...]

PR #475, opened 2018-09-27
Streamline docker setup guide

### LINKED ISSUE #442: Update docker setup docs
As reported by @LyXman in #439, the docker setup docs needs to be updated with `MIX_ENV=prod` while running migration and seeding, as well as a `BASE_URL`.

unnawut:
Closes #442

# Overview

This PR streamlines the docker setup guide as per feedbacks from the various workshops, #439 and #442.

# Changes

- Updated docker setup guide

# Implementation Details

Before, the guide branches off at step one for the user to pick between docker compose and plain docker, then they converge back in step 2. This is difficult to follow so now the option branches off right at the beginning and there's no convergence in the middle.

The commands have so been updated with more default values to make the setup easier.

# Usage

Follow the commands in the changed guide.

# Impact

Changes to user documentation only.


unnawut:
Will recheck the commands again once more before asking for review.

PR #479, opened 2018-10-01
Add full_name and calling_name to user

### LINKED ISSUE #429: Add optional name fields to users
Add the following fields to users:

- `full_name`
- `short_name` (What should we call you?)

unnawut:
Closes #429

# Overview

Adds full_name and display_name to User schema.

# Changes

- Add a migration for two new user fields: `full_name` and `display_name`
- Updated API specs, serializers, tests

# Implementation Details

This is a straight forward addition of two new database fields.

# Usage

`/user.create` or `/user.update` with the two new fields should be successful.

# Impact

New fields added to API requests/responses. DB migration added.


unnawut:
@T-Dnzt Done. Please review again.

PR #486, opened 2018-10-05
Introduce Overlay to improve preloading/filtering

### LINKED ISSUE #464: Reformat the way preloaded fields are handled

T-Dnzt:
Issue/Task Number: #464
closes #464

# Overview

This PR introduces a new concept, overlays. They are simple modules that contain the metadata for schemas (filter fields, auto-preload, search, sort, etc.)

# Changes

- This PR should fix all N+1 problem we might have by removing all preloading from serializers and instead doing it entirely in a new module, the `Orchestrator`. The `Orchestrator` is a module used for preloading, filtering, search, sorting and pagination.
- Overlays have been added for all schemas.
- Set up advanced filtering for all schemas through the `Orchestrator` and the overlays.
- All controllers have been refactored to use the `Orchestrator` and the overlays.
- A small change has been made to prevent secondary wallets from being created for users since they are not handled properly by the client API.
- Disable warnings as errors for test ENV.

# To Do in future PRs

- Refactor and unify the response/error handling in all controllers
- Handle mapped fields in the `MatchParser` module


apps/ewallet/lib/ewallet/web/orchestrator.ex
```diff
@@ -0,0 +1,67 @@
+defmodule EWallet.Web.Orchestrator do
+  @moduledoc """
+  This module should be used in every controller to deal with searching, filtering,
```
unnawut: I suggest adding a bit more description of what the module does?

```ex
@moduledoc """
  This module orchestrates the request attributes into search, filter, preload and sort queries, 
  so that this set of features can be easily and consistently applied to controllers.

  This module should be used in every controller to deal with searching, filtering,
  preloading and sorting.
"""
```

apps/ewallet/lib/ewallet/web/orchestrator.ex
```diff
@@ -0,0 +1,67 @@
...
+    query
+    |> preload_to_query(overlay, attrs)
+    |> MatchAllParser.to_query(attrs, overlay.filter_fields())
+    |> SearchParser.to_query(attrs, overlay.search_fields, default_mapped_fields())
+    |> SortParser.to_query(attrs, overlay.sort_fields, default_mapped_fields())
+  end
+
+  def all(records, overlay, attrs \\ nil)
+
+  def all(records, _overlay, %{"preload" => preload}) when is_map(preload) do
```
unnawut: Can we rename to `preload_all`? Reading it the first time, I was thinking if it's similar to [`Repo.all/2`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:all/2), but it's about preloading.

apps/ewallet/lib/ewallet/web/orchestrator.ex
```diff
@@ -0,0 +1,67 @@
...
+
+  def all(records, _overlay, %{"preload" => preload}) when is_list(preload) do
+    Preloader.preload_all(records, preload)
+  end
+
+  def all(records, overlay, _attrs) do
+    Preloader.preload_all(records, overlay.default_preload_assocs())
+  end
+
+  def one(record, overlay, attrs \\ nil)
```
unnawut: Can we rename to `preload_one`? Reading it the first time, I was thinking if it's similar to [`Repo.one/2`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:one/2), but it's about preloading.

apps/admin_api/lib/admin_api/v1/controllers/transaction_consumption_controller.ex
```diff
@@ -1,10 +1,9 @@
 defmodule AdminAPI.V1.TransactionConsumptionController do
   use AdminAPI, :controller
   import AdminAPI.V1.ErrorHandler
-  @behaviour EWallet.Web.Embedder
   alias AdminAPI.V1.AccountHelper
   alias EWallet.TransactionConsumptionPolicy
-  alias EWallet.Web.{Embedder, Paginator, Preloader, SearchParser, SortParser}
```
unnawut: Since we no longer use `Embedder` anywhere, can we delete the orphan module?

apps/ewallet/lib/ewallet/web/orchestrator.ex
T-Dnzt: It "orchestrates" `all` or `one` - currently, it's true it only preloads, but we might add more things in the future and I don't want the controllers to necessarily know what's going on in the orchestrator. That's why I call it that way :)

unnawut: Thanks. Wasn't aware that it will expand so the current naming is fine to me.

PR #493, opened 2018-10-08
Add Settings system

### LINKED ISSUE #436: Add settings system to allow provider to manage the configuration of their eWallets

T-Dnzt:
Issue/Task Number: #436
closes #436

# Overview

This PR adds a settings system to the eWallet. What started simple ended up being a very complex feature to many unexpected problems. In order to manage settings/configuration of the eWallet in a way that users can update it without updating the ENV and restarting, a new sub-application had to be created: `ewallet_config`.

# Changes

- Created `ewallet_config`
- Added the `EWalletConfig.Config` that acts as an entry point to the configuration system.
- Turn that config module into a GenServer in order to allow other applications to register themselves as settings users. See implementation details for flow.
- Added tests in `ewallet_config`
- Added test helpers to allow the Config GenServer to be used in the tests of other sub applications

# Implementation Details

## Flow

Due to some of our libraries requiring their own Application env (`arc`, etc.), we needed a way to keep using the famous `Application.get_env()`. In order to do that, the implemented system works in the following way:

1. Seeds have to be run first, to insert the current set of 19 settings used in the eWallet
2. `ewallet_config` boots up and starts a GenServer allowing other apps to register themselves
3. Other apps register themselves, specifying which settings they want to use
4. The Application envs of those apps are set using the values from the DB
5. When a setting is updated, the GenServer loops through the registered apps and refresh all their configuration. The current implementation doesn't optimize what gets updated, all apps get all their configuration refreshed.
6. To prevent frequent reloads, it is possible to update multiple settings at once.

## Settings

Settings are a module interface to the `StoredSetting` Ecto schema. This interface was needed to allow any data to be stored inside a `data` field (or an encrypted one `encrypted_data`) which is a JSON field.

From outside, it only looks like you're passing `value` (a property of the `Setting` struct), but in the background, it gets translated into the format supported in `StoredSetting`. 

A stored setting has the following fields:

- `uuid`
- `id`
- `key`: the human-readable identifier of the setting
- `data`: the unencrypted data of the setting (stored as `{value: "something"}`)
- `encrypted_data`: the encrypted data of the setting (stored as `{value: "something"}`)
- `type`: the type of the data stored in `data`/`encrypted_data`. The following types are available:    "string", "integer", "map",  "boolean", "array"
- `description`: a description of the setting
- `options`: a list of potential values for the setting
- `parent`: the `key` of the parent setting 
- `parent_value`: the `value` of the parent setting for this one to be enabled
- `secret`: if the setting data should be encrypted or not
- `position`: the position of the setting to maintain order and prevent chaos

Here's an example of a setting:

```
%{
      key: "email_adapter",
      value: "smtp",
      type: "select",
      options: ["smtp", "local", "test"],
      description:
        "When set to local, a local email adapter will be used. Perfect for testing and development."
}
```

# Usage (example with the `ewallet_db` sub app)

1. Add the list of settings in `config.exs` for the sub app

```
config :ewallet_db,
  ecto_repos: [EWalletDB.Repo],
  env: Mix.env(),
  schemas_to_audit_types: audits,
  audit_types_to_schemas: Enum.into(audits, %{}, fn {key, value} -> {value, key} end),
  settings: [
    :base_url,
    :min_password_length,
    :file_storage_adapter,
    :aws_bucket,
    :aws_region,
    :aws_access_key_id,
    :aws_secret_access_key,
    :gcs_bucket,
    :gcs_credentials
  ]
```

2. Update `application.ex` to register the app

```
settings = Application.get_env(:ewallet, :settings)
Config.register_and_load(:ewallet, settings)
```

3. Done! Anytime, a setting is updated in the DB, the Application env will be reloaded.

# Impact

We need to do a good amount of testing in staging to ensure the settings are properly refreshed in all nodes. Other than that, `Application.get_env` can still be used to retrieve config in other sub apps.

Seeds need to be run before this feature can be used. The setting seeding is ran as part of the usual `mix seed` command but can also be ran specifically with `mix seed --settings`. A prompt will ask the user for the `base_url` value to set - this will be skipped when specifying the auto-yes option `-y` (it will use the default value).

# Changes coming to this PR

- [x] Removing the `select` type, instead the `options` field will define if the value has to match some specific values
- [x] Adding validators for the type and the select values
- [x] Changing the options to an array in the database

# To Do in future PRs

- [ ] Add the endpoints to update those settings
- [ ] Add a test testing that config in all nodes get reloaded properly. I currently have the following test that doesn't work - it seems the nodes aren't connecting to the same database, and checking out the repo for each node doesn't seem to help.

```
defmodule EWalletConfig.MultiNodeTest do
  use ExUnit.Case, async: false
  alias EWalletConfig.{Config, Repo, ConfigTestHelper}
  alias Ecto.Adapters.SQL.Sandbox

  describe "reload_config/1" do
    test "reloads all settings for all nodes" do
      Sandbox.checkout(Repo)

      ConfigTestHelper.spawn([:test1, :test2, :test3])
      nodes = [Node.self() | Node.list()]

      # Not sure if this is needed, doesn't work anyway
      # Enum.each(nodes, fn node ->
      #  :rpc.block_call(node, Sandbox, :checkout, [Repo])
      # end)

      Sandbox.mode(Repo, {:shared, self()})

      Config.insert(%{key: "my_setting", value: "some_value", type: "string"})

      Enum.each(nodes, fn node ->
        :ok = Config.register_and_load(:my_app, [:my_setting], {Config, node})
        value = :rpc.block_call(node, Application, :get_env, [:my_app, :my_setting])
        assert value == "some_value"
      end)

      Config.update("my_setting", %{value: "new_value"})

      Enum.each(nodes, fn node ->
        value = :rpc.block_call(node, Application, :get_env, [:my_app, :my_setting])
        assert value == "new_value"
      end)
    end
  end
```


apps/ewallet_config/test/ewallet_config/utils/types/wallet_address_test.exs
```diff
@@ -43,25 +43,26 @@ defmodule EWalletDB.Types.WalletAddressTest do
...
-    test "uses the given wallet address if provided" do
-      {:ok, wallet} =
-        :wallet
-        |> params_for(address: "test-1234-5678-9012")
-        |> Wallet.insert()
-
-      assert wallet.address == "test123456789012"
-    end
-  end
+  #
```
unnawut: Should we re-enable these tests?

apps/ewallet_db/lib/ewallet_db/mint.ex
```diff
@@ -15,7 +15,7 @@ defmodule EWalletDB.Mint do
     external_id(prefix: "mnt_")
 
     field(:description, :string)
-    field(:amount, EWalletDB.Types.Integer)
+    field(:amount, EWalletConfig.Types.Integer)
```
unnawut: Should types be part of `EWalletDB` than `EWalletConfig`? Especially when it's used to define the schema's field type.

apps/ewallet_db/lib/ewallet_db/validator.ex
```diff
@@ -5,135 +5,6 @@ defmodule EWalletDB.Validator do
...
-    case validate_password(password) do
-      {:ok, _} ->
-        changeset
-
-      {:error, :password_too_short, data} ->
-        Changeset.add_error(changeset, key, "must be #{data[:min_length]} characters or more")
-    end
-  end
-
   def validate_from_wallet_identifier(changeset) do
```
unnawut: Just wondering why this function did not get moved to `EWalletConfig` like the other?

apps/ewallet_db/priv/repo/reporters/seeds.exs
```diff
@@ -1,6 +1,8 @@
 defmodule EWalletDB.Repo.Reporters.SeedsReporter do
+  alias EWalletConfig.Config
+
   def run(writer, args) do
-    base_url = Application.get_env(:ewallet_db, :base_url) || "https://example.com"
+    base_url = Config.get("base_url", "https://example.com")
```
unnawut: I couldn't find `Config.get/2` in `EWalletConfig.Config`.

unnawut:
One final note is I'm not sure a race condition will be a problem here. Would be nice to have @sirn have a look at this PR as well.

apps/ewallet_config/lib/ewallet_config/setting.ex
```diff
@@ -0,0 +1,261 @@
...
+  end
+
+  @doc """
+  Inserts all the default settings.
+  """
+  @spec insert_all_defaults(Map.t()) :: [{:ok, %Setting{}}] | [{:error, %Changeset{}}]
+  def insert_all_defaults(overrides \\ %{}) do
+    get_default_settings()
+    |> Enum.map(fn {key, data} ->
+      case overrides[key] do
```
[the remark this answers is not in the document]
T-Dnzt: Adding it.

apps/ewallet_config/lib/ewallet_config/stored_setting.ex
```diff
@@ -0,0 +1,73 @@
...
+  schema "setting" do
+    external_id(prefix: "stg_")
+
+    field(:key, :string)
+    field(:data, :map)
+    field(:encrypted_data, EWalletConfig.Encrypted.Map)
+    field(:type, :string)
+    field(:description, :string)
+    field(:options, :map)
+    field(:parent, :string)
```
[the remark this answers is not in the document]
T-Dnzt: - `position`: Position is used to have a constant order for settings that we define. It cannot be updated and is only set at creation. We can add more settings in the seeds later and fix their positions.
- `parent` and `parent_value`: those are used to link settings together in a very simple way. No logic is actually implemented for those, it's mostly intended to be used by clients (like the admin panel) to show settings in a logical way. So if someone selects `gcs` for `file_storage`, you can show all settings that have `file_storage` as a parent were `parent_value=gcs`.

apps/ewallet_config/lib/ewallet_config/setting.ex
```diff
@@ -0,0 +1,261 @@
...
+  end
+
+  defp cast_attrs(attrs) do
+    attrs
+    |> cast_value()
+    |> cast_options()
+    |> add_position()
+  end
+
+  defp build_changeset(changeset) do
```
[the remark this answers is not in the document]
T-Dnzt: Because that first changeset is tainted, it contains details about the internal schema `StoredSetting` so we need to replace/clean the returned changeset before sending it back to the sender.

apps/ewallet_config/lib/ewallet_config/validator.ex
```diff
@@ -0,0 +1,170 @@
+defmodule EWalletConfig.Validator do
```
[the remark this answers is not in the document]
T-Dnzt: If it's part of `EWalletDB`, then we can't use those validation functions in `EWalletConfig`. `EWalletConfig` has no dependencies on any other app, but all other apps depend on it directly or indirectly. I could move back some of the functions to `EWalletDB` tho, since they are not used in the validations in `EWalletConfig` anymore.

apps/ewallet_config/priv/repo/migrations/20181008093853_add_settings.exs
```diff
@@ -0,0 +1,27 @@
...
+      add :parent_value, :string
+      add :secret, :boolean, null: false, default: false
+      add :position, :integer, null: false, unique: true
+
+      timestamps()
+    end
+
+    create unique_index(:setting, [:id])
+    create unique_index(:setting, [:key])
+    create unique_index(:setting, [:position])
```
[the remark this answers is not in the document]
T-Dnzt: Inserting settings in existing positions shouldn't really happen. If we ever need to do it though, we could write a function to update all the positions - not that big of a deal since it would only be ran once. The position logic isn't great to be honest, but I think it's good enough for now. I'd like to actually switch to a category system where each category has its set of settings.

I can remove the unique index on the position so we don't have to care much, but it doesn't mean we should start having settings with the same positions - because then, you might show the settings in a different list to the admin panel users every time you reload the page.

apps/ewallet_config/priv/repo/migrations/20181008093853_add_settings.exs
```diff
@@ -0,0 +1,27 @@
...
+      add :key, :string, null: false
+      add :data, :map
+      add :encrypted_data, :binary
+      add :type, :string, null: false
+      add :description, :string
+      add :options, :map
+      add :parent, :string
+      add :parent_value, :string
+      add :secret, :boolean, null: false, default: false
+      add :position, :integer, null: false, unique: true
```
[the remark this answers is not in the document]
T-Dnzt: 👌

apps/ewallet_config/test/ewallet_config/setting_test.exs
```diff
@@ -0,0 +1,409 @@
...
+    test "inserts a setting with an array value" do
+      attrs = %{key: "array_key", value: ["a", "b", "c"], type: "array"}
+      {res, setting} = Setting.insert(attrs)
+
+      assert res == :ok
+      assert setting.uuid != nil
+      assert setting.value == ["a", "b", "c"]
+    end
+
+    test "inserts a CRON schedule" do
```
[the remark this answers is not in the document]
T-Dnzt: I finally removed the settings for the scheduler because it wasn't working, but didn't remove the test :P

apps/ewallet_db/lib/ewallet_db/mint.ex
T-Dnzt: I don't think it should, and this would actually allow us to clean up the shared ewallet_db / local_ledger_db stuff. The config is lower than the db apps and can be used by any other sub app.

apps/ewallet_db/lib/ewallet_db/validator.ex
T-Dnzt: Because it uses `EWalletDB.Wallet` :)

apps/ewallet_config/test/ewallet_config/utils/types/wallet_address_test.exs
T-Dnzt: No they have to be moved to another test file actually, or be deleted completely. Since they interact with the database and assume this module is being used a part of a schema, it shouldn't be here. This is testing the integration of the `WalletAddress`, not the `WalletAddress` itself.

apps/ewallet_db/priv/repo/reporters/seeds.exs
T-Dnzt: It's the main one -> https://github.com/omisego/ewallet/blob/6b30ffd209baab97211507f6b6b5455511c32f5f/apps/ewallet_config/lib/ewallet_config/config.ex#L121

T-Dnzt:
@unnawut That's a good point about race conditions. I'm gonna fix it.

apps/ewallet_config/lib/ewallet_config/stored_setting.ex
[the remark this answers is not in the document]
unnawut: Could you add this as a code comment? I think this will be easily lost

apps/ewallet_config/lib/ewallet_config/validator.ex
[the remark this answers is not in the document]
unnawut: Moving back would be nice I think. I feel `EWalletDB` having to depend on validators from `EWalletConfig` is a bit counterintuitive.

Also, if the `EWalletConfig`'s validators are the same as `EWalletDB`'s, having duplicate code in this is case fine?

mederic-p:
You should update https://github.com/omisego/ewallet/blob/master/docs/design/components.md

apps/admin_api/lib/admin_api/application.ex
```diff
@@ -9,6 +9,10 @@ defmodule AdminAPI.Application do
   def start(_type, _args) do
     import Supervisor.Spec
     DeferredConfig.populate(:admin_api)
+
+    settings = Application.get_env(:admin_api, :settings)
```
sirn: Hmm, I have a feeling that `:settings` shouldn't be in config (since it's not exactly configurable). Maybe as a `@settings [...]` in here (i.e. `application.ex`) instead?

sirn: Or if we do it might make sense to use something among the line of:

```
config :admin_url,
  sender_email: {:ewallet_config, "default"},
  something_else: {:ewallet_config, nil}
```

…and iterate over the keys. In this case we can simplify `EWalletConfig.Config.register_and_load` to just:

```
EWalletConfig.Config.register_and_load(:admin_api)
```

apps/ewallet/lib/mix/tasks/omg.seed.ex
```diff
@@ -49,7 +49,11 @@ defmodule Mix.Tasks.Omg.Seed do
   #
 
   defp seed_spec([]) do
-    [{:ewallet_db, :seeds}]
+    [{:ewallet_db, :seeds}] ++ seed_spec(["--settings"])
```
sirn: I think this should be part of the default seeds na (just add it to `seed_spec([])`'s list rather than a new function to handle it). `seed_spec([])` should be the "final" function to get called.

apps/ewallet/priv/repo/report_minimum.exs
```diff
@@ -1,9 +1,10 @@
 alias EWallet.CLI
+alias EWalletConfig.Config
 alias EWalletDB.AuthToken
 
 # :prod environment does not have a default :base_url value and should not have one.
 # But we have a fallback value here so we can generate a friendly output message for seeding.
-base_url = Application.get_env(:ewallet_db, :base_url) || "https://example.com"
+base_url = Application.get_env(:ewallet, :base_url) || "https://example.com"
```
sirn: Do we still need the default `https://example.com/`? 🤔

apps/admin_api/lib/admin_api/application.ex
T-Dnzt: Hmm, but from my point of view, it is configurable - you can set which settings you need in the app for it to work, we can add more in the future or remove them later. I can't really see how it's different from other config like `generators` or the `mime_types`. 

About listing the settings, we could do that, but I don't really like the idea of having default in there. I see the settings as a global system, with one default value for all the apps, don't really want to let each app set their own default. At least, that's how I see it with the current implementation, it might evolve and requires apps to each have different default values 🤷‍♂️

apps/ewallet/lib/mix/tasks/omg.seed.ex
T-Dnzt: What I tried to do here is give the option to only run the settings seeds for all our systems that are already deployed (thought it'd be nice :D). Totally fine to change it, but not sure I follow what you mean, can you give me some code?

apps/ewallet/priv/repo/report_minimum.exs
T-Dnzt: I guess not, removing it!

apps/admin_api/lib/admin_api/application.ex
sirn: Maybe we can use `{:ewallet_config, "key_to_retrieve"}` or something (where `nil` defaults to the key)?

My issue is mainly with having `:settings` key in config, which doesn't feel exactly right to me. As in, we're defining the values to be populated into `app_env` in a different ways than how these settings are normally populated (as in DeferredConfig, or other MFA-style configurations).

Also, if we use `{:ewallet_config, ...}` we could stub it in tests without involving `ewallet_config` at all.

sirn:
Thibault <notifications@github.com> writes:

Aha. Got it. It's usually preferrable to have the final function for
these type of recursive functions to no longer call itself because it's
usually simpler to follow.

For now I think it's OK to leave it like this. Although I think we
should find a way to handle this case (and case like `--test`) in the
future, since we will need more seed types in the future for sure.

apps/admin_api/lib/admin_api/application.ex
T-Dnzt: @sirn I'm not sure why we need that `{:ewallet_config, ...` part for? Also, one of the reason it's in the settings in that format is that I'm actually using the whole list of settings in the shared test helper function  there https://github.com/omisego/ewallet/blob/436-settings-system/apps/ewallet_config/test/support/config_test_helper.ex#L15.

sirn:
Well, we already have `{:system, ...}` and `{:apply, ...}` syntax a
a result of using DeferredConfig. I think it make sense if we follow the
same syntax rather than introducing a new one.

Besides it's nicer when we want to make the configuration key use other
configuration scheme (e.g. hard-coded or using DeferredConfig), such as
in tests, or in development.

T-Dnzt:
Merging for now, I've given a try to the `{:ewallet_config, nil}` approach but it would require too many changes, let's do it in another PR.

sirn:
:ok_hand:

PR #515, opened 2018-11-09
Add configuration/settings endpoints

### LINKED ISSUE #502: Allow settings to be updated through Admin API endpoints

T-Dnzt:
Issue/Task Number: #502 
closes #502

# Overview

This PR adds two new endpoints to manage the configuration settings:

- `/configuration.get`
- `/configuration.update`

I decided to go with `configuration.*` instead of `settings.` simply because we already have a `/settings.all` endpoint that returns something in the shape of 

```
%{
  minted_tokens: []
}
```

I didn't know how to reuse that and make it fit with the settings system, so using something like `/configuration.*` made more sense. Totally up for discussion if you guys have better ideas.

# Changes

- Add `count` to the pagination object to know exactly how many records are being returned in the response (suggested by one of our mobile developers, a small change that makes it easier to parse a response and adds little overhead in the code).
- Add `/configuration.get` to retrieve the list of all settings. This actually returns a paginated, filterable and sortable list of settings. `/configuration.all` could be a better name to stay consistent.
- Add `/configuration.update` that takes a hash of settings update in the form of:

```
%{
  "setting_name_1": "setting_value_1",
  "setting_name_2": "setting_value_2",
}
```

And will return a list of insert results which can either be:

- an updated setting (`configuration_setting` type)
- an error changeset (`error` type)
- a `configuration_setting_error` (that would occur in case a setting key is not found)

# To Do

- [ ] Update Swagger


apps/admin_api/test/admin_api/v1/controllers/admin_auth/configuration_controller_test.exs
```diff
@@ -0,0 +1,111 @@
...
+               "description" =>
+                 "Invalid parameter provided. `value` must be one of 'smtp', 'local', 'test'.",
+               "messages" => %{"value" => ["value_not_allowed"]},
+               "object" => "error"
+             }
+
+      assert data["fake_setting"] == %{
+               "code" => "setting_not_found",
+               "key" => "fake_setting",
+               "object" => "configuration_setting_error"
```
mederic-p: Do we really need to have 2 different kind of `error` objects returned? If possible I'd stick with only `error` and try to remove `configuration_setting_error`

unnawut: +1

apps/ewallet_config/lib/ewallet_config/config.ex
```diff
@@ -102,9 +102,28 @@ defmodule EWalletConfig.Config do
...
+    end
   end
 
+  defp get_config_pid(%{config_pid: config_pid} = attrs),
+    do: {config_pid, Map.delete(attrs, :config_pid)}
+
+  defp get_config_pid(%{"config_pid" => config_pid} = attrs),
+    do: {config_pid, Map.delete(attrs, "config_pid")}
+
+  defp get_config_pid([{:config_pid, config_pid}] = attrs),
```
unnawut: I think this works only if the list is exactly `[{:config_pid, config_pid}]`. It won't work with `[{:some_other_key, some_other_value}, {:config_pid, config_pid}]`, etc.

Probably need to use `Keyword.fetch/3` or `Keyword.get/3` within the function implementation instead.

apps/ewallet_config/lib/ewallet_config/config.ex
```diff
@@ -102,9 +102,28 @@ defmodule EWalletConfig.Config do
...
+    do: {config_pid, Keyword.delete(attrs, :configpid)}
```
unnawut: `:configpid` -> `:config_pid`?

apps/admin_api/priv/spec.yaml
```diff
@@ -3345,40 +3365,144 @@ components:
             expired_at: null
 
     ######################################
-    #           SETTING SCHEMAS          #
+    #       CONFIGURATION SCHEMAS        #
     ######################################
-    SettingResponse:
-      description: "The response schema for settings"
+    SettingSchema:
```
unnawut: Since we're using "configuration" now, should this be changed to `ConfigurationSchema`?

Also `SettingResponse` -> `ConfigurationResponse` a bit down below.

T-Dnzt: I wish it was just settings, but it was taken, so instead we send back a configuration composed of a set of settings, that's why you see both settings and configuration in the spec depending on the level.

Issue #490, opened 2018-10-08
Implement permission matrix


T-Dnzt:
**Current behavior:**
- Each controller contains a private `permit/3` function
  - `permit/3` takes in the `action` to be authorized, the `params` to use to figure out the requester, and the 3rd argument as the target to authorize with.
  - `permit/3`, then in turn, calls `Bodyguard.permit/4` with an `EWallet.Policies.*Policy` to try authorize with the provided info.
- The `EWallet.Policies.*Policy` is where the actual authorization conditions are specified. It takes in the action and the information about the requester (e.g. user identifier, access key, etc.) and maps to the actual permission. Currently all this is hardcoded.
- The return value is either `:ok` or `{:error, reason}` with the default error reason being `:unauthorized` if not specified.

**Extra implementation needed:**
- [x] Add `/role.all` - retrieve a list of roles
- [x] Add `/role.get` - retrieve a specific role
- [x] Add `/role.create` - create a role
- [x] Add `/role.update` - update a role
- [ ] A database table called `permission` that maps each role to each permission
  - e.g. `role_id`, `key_id`, `entity`, `action`, `allowed`
  - `role_id` and `key_id` are mutually exclusive, a permission could be granted to either a role or a key
- [ ] Some mechanism to populate new permissions to DB and prune deleted ones (probably during migration)
- [ ] Update `EWallet.Policies.*Policy` to query the permissions from DB rather than hardcode.
- [ ] Add `/permission.all` - retrieve a list of all permissions
- [ ] Add `/role.get_permissions` - retrieve a list of permissions for a specific role
- [ ] Add `/role.update_permissions` - update permissions for a specific role
- [ ] Caching through ETS so permission lookups do not require DB requests every single time.

PR #519, opened 2018-11-13
Add /role.all, get, create, update, delete to Admin API

unnawut:
Issue/Task Number: #490
Required by #490

# Overview

This PR implements the role endpoints (all, get, create, update, delete) required to continue with the actual permissions in #490.

This is a separate PR from the rest of the permissions to make the review easier.

# Changes

- Added controllers and supporting modules for /role.all, /role.get, /role.create, /role.update, /role.delete
- Refactored `Role.get_by_name/1` into `Role.get_by/2`
- Added external ID and soft deletion to `Role` schema
- Added OpenAPI specs for the new endpoints

# Implementation Details

Straightforward adding of endpoints to retrieve and manage roles. The `Role` schema did not have external ID nor soft deletion before, but to keep the endpoints and their usage consistent, it makes sense to add them so implementers can expect consistent behavior across the eWallet.

# Usage

Retrieve all roles:

```sh
curl 10.5.10.10:4000/api/admin/role.all \
-H "Authorization: OMGAdmin <truncated>" \
-d '{
}' \
-v -w "\n"
```

Update a role:

```sh
curl 10.5.10.10:4000/api/admin/role.update \
-H "Authorization: OMGAdmin <truncated>" \
-d '{
  "id": "rol_01cw625zm2xmny5ee3mwt9j8ky",
  "name": "new_name"
}' \
-v -w "\n"
```

# Impact

DB migration and new API endpoints


apps/admin_api/lib/admin_api/v1/controllers/role_controller.ex
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  def update(conn, _), do: handle_error(conn, :invalid_parameter)
+
+  @doc """
+  Soft-deletes an existing role by its id.
+  """
+  @spec delete(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def delete(conn, %{"id" => id} = attrs) do
+    with %Role{} = role <- Role.get(id) || {:error, :role_id_not_found},
```
mederic-p: Shouldn't we have a `permit` here too?

unnawut: Updated. Please check again.

apps/admin_api/test/admin_api/v1/controllers/admin_auth/role_controller_test.exs
```diff
@@ -0,0 +1,216 @@
...
+
+      response = admin_user_request("/role.all", attrs)
+      roles = response["data"]["data"]
+
+      assert response["success"]
+      assert Enum.count(roles) == 3
+      assert Enum.at(roles, 0)["name"] == "matched_3"
+      assert Enum.at(roles, 1)["name"] == "matched_2"
+      assert Enum.at(roles, 2)["name"] == "matched_1"
+    end
```
T-Dnzt: We should start adding a test that uses the match parsers for all the `.all` endpoints.

apps/ewallet/lib/ewallet/policies/role_policy.ex
```diff
@@ -0,0 +1,23 @@
+defmodule EWallet.RolePolicy do
+  @moduledoc """
+  The authorization policy for roles.
+  """
+  @behaviour Bodyguard.Policy
+  alias EWalletDB.{Account, User}
+
+  # Any user can get a role
+  def authorize(:all, _user_or_key, _role_id), do: true
```
T-Dnzt: Let's be more explicit here and use `admin_user:` or `key: `

apps/ewallet/lib/ewallet/web/v1/serializers/role_serializer.ex
```diff
@@ -0,0 +1,44 @@
...
+
+  def serialize(%Role{} = role) do
+    %{
+      object: "role",
+      id: role.id,
+      name: role.name,
+      priority: role.priority,
+      display_name: role.display_name,
+      user_ids: UserSerializer.serialize(role.users, :id),
+      users: UserSerializer.serialize(role.users),
```
T-Dnzt: Do we really want this? I don't think we should send back users here na khrap.

apps/admin_api/test/admin_api/v1/controllers/admin_auth/role_controller_test.exs
unnawut: Updated. Please check.

apps/ewallet/lib/ewallet/policies/role_policy.ex
unnawut: Updated. Please check.

apps/ewallet/lib/ewallet/web/v1/serializers/role_serializer.ex
unnawut: Updated. Please check.

apps/admin_api/test/support/conn_case.ex
```diff
@@ -309,4 +315,105 @@ defmodule AdminAPI.ConnCase do
     insert_list(num_remaining, factory_name, attrs)
     Repo.all(schema)
   end
+
+  @doc """
+  Tests that the specified endpoint supports 'match_any' filtering.
+  """
+  defmacro test_supports_match_any(endpoint, auth_type, factory, field, opts \\ []) do
```
T-Dnzt: 👍

PR #553, opened 2018-11-23
Reset password feature for standalone eWallet users

### LINKED ISSUE #432: Standalone forget password feature
This issue adds the forget password feature which is not part of the current standalone wallet issue (#357).

unnawut:
Issue/Task Number: #432
Closes #432

# Overview

This PR adds the remaining reset password feature for standalone users, which was deferred from #357.

# Changes

- Refactored the admin's reset password feature from `AdminAPI` into `EWallet` namespace so it can be shared with the `EWalletAPI`.
- Added endpoints and controllers for `/api/client/user.reset_password` and `/api/client/user.update_password`

# Implementation Details

Because the feature is the same as what's already present in `/api/admin/admin.reset_password` and `/api/admin/admin.update_password`. This is a striaghtforward refactor of the AdminAPI controllers into a shared `EWallet.ResetPasswordGate` and use this common gate for the two APIs.

# Usage

**Initiate the reset password flow:**

```sh
curl 10.5.10.10:4000/api/client/user.reset_password \
-d '{
  "email": "testinvite-dfgdg@example.com",
  "redirect_url": "http://localhost:4000/reset_password?email={email}&token={token}"
}' \
-v -w "\n" | jq
```

**Update the password:**

```sh
curl 10.5.10.10:4000/api/client/user.update_password \
-d '{
  "email": "testinvite-dfgdg@example.com",
  "token": "Qn153G20eWdUnJtSqsb4BjxnbOaCDQpHfN3qwJ7c24o",
  "password": "new.password",
  "password_confirmation": "new.password"
}' \
-v -w "\n" | jq
```

# Impact

New API endpoints: `/api/client/user.reset_password` and `/api/client/user.update_password`. No changes to the DB schema.


apps/ewallet/lib/ewallet/gates/reset_password_gate.ex
```diff
@@ -0,0 +1,69 @@
...
+    end
+  end
+
+  # Private functions
+
+  defp get_user_by_email(nil), do: {:error, :user_email_not_found}
+
+  defp get_user_by_email(email) do
+    case User.get_by_email(email) do
+      nil -> {:error, :user_email_not_found}
```
T-Dnzt: I don't think we should return that error when requesting a reset. This is letting know a potential attacker that an email isn't used in the system. When he actually finds an email used, he won't receive any error and will have been able to guess private information. I think we should always return `:ok` when a reset is done with a message like `A reset email has been sent to the given address if it was found in our system.`. Thoughts?

apps/ewallet/lib/ewallet/gates/reset_password_gate.ex
```diff
@@ -0,0 +1,69 @@
...
+  """
+  alias EWalletDB.{ForgetPasswordRequest, User}
+
+  @doc """
+  Creates a reset password reset.
+  """
+  @spec request(map()) :: %ForgetPasswordRequest{}
+  def request(email) do
+    with {:ok, user} <- get_user_by_email(email),
+         {_, _} <- ForgetPasswordRequest.disable_all_for(user),
```
T-Dnzt: Now I'm also wondering about the `disable_all`. Someone could spam reset an email address, preventing a real user from doing a password reset. I'm thinking we can add a lifetime to requests instead? We could add a setting for the request lifetime.

unnawut:
Remaining:
- Add settings for reset password request lifetime, default to 10 minutes
- Set `expires_at` on the generated forget password request based on the settings
- Double check the `expires_at` before allowing the password update
- Disables the request & set `used_at` after used
- Create a scheduled job to expire the request after `expires_at` is passed

unnawut:
@T-Dnzt Updated. Please check again.

mederic-p:
Can you update the open api spec? ;)

PR #569, opened 2018-11-28
Add version informations in status responses

### LINKED ISSUE #562: Add version number in the root endpoint
# Subject of the issue

I would like to be able to retrieve the API version from the root endpoint

# Steps to reproduce

Access http://localhost:4000/api/admin

# Expected behavior

There is a `version` parameter in the response of each request, it would also be good to have it in the root endpoint, something such as : 
```
{"success":true,"version":"1"}
```

This could be useful for 3rd party libraries implementing this SDK, to be able to know from the get go if the API base URL is compatible with the client's implementation

# Actual behavior

root endpoint returns :
```
{"success":true}
```

mederic-p:
Issue/Task Number: 562
Closes #562 

# Overview

This PR adds additional info mostly about versions to the different status endpoints.

# Changes

- `/` -> added `ewallet_version: "1.1.0"`
- `/api/admin` and `/api/client` -> added `"ewallet_version": "1.1.0"` and `"api_versions":[{"name":"v1","media_type":"application/vnd.omisego.v1+json"}]}`
- updated `mix.exs` files in all sub-apps to inherit version from application variable

# Implementation Details

Added a new application variable for the `:ewallet` -> `:version`

# Usage

Check the `/`, `/api/admin` and `/api/client` jsons.

# Impact

Additional info in json responses.


config/config.exs
```diff
@@ -9,7 +9,7 @@ use Mix.Config
 # back to each application for organization purposes.
 import_config "../apps/*/config/config.exs"
 
-config :ewallet, root: File.cwd!()
+config :ewallet, root: File.cwd!(), version: "1.1.0"
```
unnawut: @sirn It's possible to dynamically get the version from the `Mixfile`'s `project.version` but I guess that doesn't work with distillery. Any suggestions?

sirn:
Maybe do that in dev.exs and leave the value as "prerelease" or something in prod.exs. I have a script that can be used to update versions so maybe I can use that.

unnawut:
Nevermind me, it seems the Mixfile is able to take the version config successfully.

PR #575, opened 2018-12-04
Configuration system for admin panel

jarindr:
Issue/Task Number: #557 

# Overview
This PR add configuration system page to admin panel.

# Changes
- Add configuration system to admin panel.
- rename endpoint from `.get`  to `.all` for consistency.
- rename module ConfigurationSerializer.
- refactor redux store loading state.
- change how configuration endpoint works, from paginated to returning all data.
- remove deprecated function like `Ecto.date` and fix test case.

# Implementation Details
- the old redux store loading state for each node was separated before for flat tree, but we realize it makes tree too ugly and hard to maintain so we changed back to a single state tree for all loading state.

# Usage
Go to configuration page and test around.

# Impact
- Deploy as usual, anyone testing this branch please clean your compiled code and db first to avoid bugs.

screen shot 2018-12-11 at 5 03 35 pm


mederic-p:
I think it'd be nice to have a disclaimer on the top of the page telling that some of these setting are sensitive and shouldn't be modified unless you know what you are doing

apps/admin_api/lib/admin_api/v1/controllers/configuration_controller.ex
```diff
@@ -2,13 +2,14 @@ defmodule AdminAPI.V1.ConfigurationController do
   use AdminAPI, :controller
   import AdminAPI.V1.ErrorHandler
 
-  alias EWallet.Web.{Orchestrator, Originator, V1.SettingOverlay}
+  alias EWallet.Web.{Orchestrator, Originator, V1.ConfigurationOverlay}
   alias EWalletConfig.{Config, Repo}
 
   def get(conn, attrs) do
```
unnawut: Maybe change `def get(conn, attrs) do` to `def all(conn, attrs) do` as well?

apps/admin_api/test/admin_api/v1/controllers/admin_auth/configuration_controller_test.exs
```diff
@@ -2,45 +2,20 @@ defmodule AdminAPI.V1.AdminAuth.ConfigurationControllerTest do
...
-
-      first_setting = Enum.at(response["data"]["data"], 0)
-      last_setting = Enum.at(response["data"]["data"], -1)
-
-      assert first_setting["key"] == "base_url"
-      assert first_setting["position"] == default_settings["base_url"].position
-
-      assert last_setting["key"] == "aws_secret_access_key"
-      assert last_setting["position"] == default_settings["aws_secret_access_key"].position
+      assert length(response["data"]["data"]) == 19
```
unnawut: Just so we don't need to keep updating this value, use this instead?

```ex
assert length(response["data"]["data"]) == length(Application.get_env(:ewallet_config, :default_settings))
```

apps/ewallet_config/config/config.exs
```diff
@@ -12,12 +12,19 @@ config :ewallet_config,
   },
   default_settings: %{
     # Global Settings
-    "base_url" => %{key: "base_url", value: "", type: "string", position: 100},
+    "base_url" => %{
+      key: "base_url",
+      value: "",
+      type: "string",
+      position: 100,
+      description: "The domain name where this eWallet is accessible."
```
unnawut: Can we rephrase to avoid the word "domain name"? Because the settings actually need a URL. E.g.

> The URL where the base of the eWallet is accessible from.

apps/ewallet_config/config/config.exs
```diff
@@ -12,12 +12,19 @@ config :ewallet_config,
...
+    },
     "redirect_url_prefixes" => %{
       key: "redirect_url_prefixes",
       value: [],
       type: "array",
-      position: 101
+      position: 101,
+      description: "The whitelisted domain names the eWallet can redirect to (for password resets, etc.)."
```
unnawut: Can we rephrase to avoid the word "domain name"? Because the settings actually need a URL. E.g.

> The URL prefixes where the eWallet are allowed to redirect to (for password resets, etc.)

apps/ewallet_config/mix.exs
```diff
@@ -45,8 +45,8 @@ defmodule EWalletConfig.MixProject do
       {:poison, "~> 3.1"},
       {:deferred_config, "~> 0.1.0"},
       {:cloak, "~> 0.7.0-alpha"},
-      {:arc, "~> 0.8.0"},
-      {:arc_ecto, "~> 0.7.0"},
+      {:arc, "~> 0.11.0"},
+      {:arc_ecto, "~> 0.11.1", github: "omisego/arc_ecto"},
```
unnawut: `"~> 0.11.1"` doesn't have any effect if `github:` is provided.

apps/ewallet_db/mix.exs
```diff
@@ -52,15 +52,16 @@ defmodule EWalletDB.Mixfile do
       {:bcrypt_elixir, "~> 1.0"},
       {:cloak, "~> 0.7.0-alpha"},
       {:plug, "~> 1.0"},
-      {:arc, "~> 0.8.0"},
-      {:arc_ecto, "~> 0.7.0"},
+      {:arc, "~> 0.11.0"},
+      {:arc_ecto, "~> 0.11.1", github: "omisego/arc_ecto"},
```
unnawut: `"~> 0.11.1"` doesn't have any effect if `github:` is provided.

PR #582, opened 2018-12-07
Make users global in admin panel

### LINKED ISSUE #530: Allow all users data to be retrieved from all accounts
In order to potentially show the users data when listing wallets, transactions, etc, we need a new flag titled `include_users_data` that will work with `owned` to provide exactly what the developers need.

T-Dnzt:
Issue/Task Number: #530 
closes #530

# Overview

This PR changes users from being scoped to being global. All accounts can see all users (but only administrators in accounts which have links with users can update them).

# Changes

- Return user data in account transactions/wallets/requests/consumptions endpoints
- Change the admin panel to use `/user.all` to list all users.


apps/admin_api/test/admin_api/v1/controllers/provider_auth/transaction_controller_test.exs
```diff
@@ -289,7 +289,8 @@ defmodule AdminAPI.V1.ProviderAuth.TransactionControllerTest do
           "per_page" => 50
         })
 
-      assert length(response["data"]["data"]) == 13
+      # 13 account transactions + 2 user transactions
```
unnawut: Love the comment

PR #585, opened 2018-12-11
Report internal server errors to Sentry

### LINKED ISSUE #358: Report to Sentry when the ErrorHandler returns an internal server error
Currently, the `ErrorHandler` can return an `internal_server_error` for errors that are not handled properly.

Since this behavior is abnormal, it should be reported to Sentry.

unnawut:
Issue/Task Number: #358 
Closes #358

# Overview

This PR adds the reporting of internal server errors to Sentry.

# Changes

- Updated `:sentry` dependency to version 6.4
- Added `use Sentry.Phoenix.Endpoint` to `AdminAPI.Endpoint` and `EWalletAPI.Endpoint`
- Add tests to check that reports are sent to Sentry

# Implementation Details

This is a straightforward implementation according to [`sentry-elixir`'s docs](https://hexdocs.pm/sentry/6.4.2/readme.html#setup-with-plug-or-phoenix).

The dependency `:cowboy` is changed to `:plug_cowboy` as suggested by `:cowboy` itself.

# Usage

Making a request to `/api/admin/status.server_error` or `/api/client/status.server_error` should file a report to the configured Sentry server.

If the response returned is not in json format, this is likely because you are on dev environment with error debugging turned on. To see the proper json response, change to `debug_errors: false` in `config/dev.exs` (for both `admin_api` and `ewallet_api` apps). Also don't forget to clean + compile after the config changes.

# Impact

No changes to the API specs or DB schema.


unnawut:
Remaining: Move /status.server_error behind auth

unnawut:
Remaining:

- [ ] Check that invalid error codes are also handled e.g.
  ```ex
  handle_error(conn, :unkown_error_that_will_send_something_to_sentry)
  ```

PR #586, opened 2018-12-12
Add AppSignal support

### LINKED ISSUE #558: Add optional support for AppSignal

unnawut:
Issue/Task Number: #558 
Closes #558 

# Overview

This PR adds AppSignal support to the eWallet.

# Changes

- Add `:appsignal` dependency
- Add configurations for AppSignal

# Implementation Details

I originally tried to add `:appsignal` as an optional dependency, but since it needed to be integrated into the endpoints, templates, etc. (e.g. `use Appsignal.Phoenix` in `Endpoint`), it cannot be an optional dependency.

# Usage

Set the following environment variable and AppSignal will be enabled automatically.

```sh
$ export APPSIGNAL_PUSH_API_KEY=<your_appsignal_api_key>
```

# Impact

No changes to the DB schema or API specs.


unnawut:
Remaining: Looking into adding a facade/decorator to uncouple Appsignal from the main apps.

unnawut:
Uncoupling the features will be a heavy task and so will be considered as a separate issue, possibly as a generic plugin system.

jbunce:
Hey @unnawut , thanks for helping out with this. Is it possible with this implementation to add custom instrumenters like in https://docs.appsignal.com/ruby/instrumentation/instrumentation.html while keeping the code base open source friendly?

unnawut:
> Is it possible with this implementation to add custom instrumenters like in https://docs.appsignal.com/ruby/instrumentation/instrumentation.html while keeping the code base open source friendly?

Yeah I think there are a couple of ways to do this. It's a large feature by itself so I'll tackle it in a separate issue/PR, but here are my mental notes:

- As far as I know, there's no dynamic library loading in Elixir, so the code for whatever we need, be it optional or not, needs to be bundled with the app.
- However, we could manage so that the 3rd party code never touches runtime unless specifically enabled. This can be done through strictly manually calling [`Application.ensure_all_started/2`](https://hexdocs.pm/elixir/Application.html#ensure_all_started/2), not including it in the default supervision tree. And all 3rd-party touching code gets called from the decorators/helper functions below.
- We could do similar to how `appsignal-elixir` go about with [function decorators](https://docs.appsignal.com/elixir/instrumentation/instrumentation.html#function-decorators) and [instrumentation helper functions](https://docs.appsignal.com/elixir/instrumentation/instrumentation.html#instrumentation-helper-functions). But instead of hard coding to AppSignal, we create own decorators/helper functions that determine whether to trigger AppSignal, another service(s), or none.
- Or go with something like [`octobus/event_bus`](https://github.com/otobus/event_bus) but it looks overly complex / unidiomatic to me.

unnawut:
I'll open a separate issue to track this.

PR #589, opened 2018-12-14
Clean up Settings documentation & workaround for SMTP_USER env variable

unnawut:
Issue/Task Number: #578

# Overview

This PR update the documentation to reflect the current environment variables and Settings in use.

This PR also adds a workaround where both `SMTP_USER` and `SMTP_USERNAME` environment variables were used.

# Changes

- Updated markdown documentations
- Add support for migrating `SMTP_USER` in `mix omg.migrate.settings`

# Implementation Details

- Documentation: straightforward moving of the obsoleted env variables into a new Settings section.

- `SMTP_USER` env var migration:  The `SMTP_USER` is not used as of late but was a valid setting in v1.0, so it's important to migrate that value otherwise upgrading from v1.0 will break. This migration should be safe enough as the newer `SMTP_USERNAME` value always takes precedence over `SMTP_USER`. This was found when trying to migrate the settings on staging environment.

# Usage

See `/docs/setup/advanced/settings.md` and try `mix omg.migrate.settings` with `SMTP_USER` set. The migration plan should indicate its migration towards the same `smtp_username` settings.

If both `SMTP_USER` and `SMTP_USERNAME` are set, the `SMTP_USERNAME` will always be migrated after `SMTP_USER`.

# Impact

No changes to DB schema or API specs.


docs/setup/advanced/settings.md
```diff
@@ -0,0 +1,76 @@
...
+## General
+
+Below are the general settings needed for the eWallet to run.
+
+- `base_url`: The base to use when building URLs.
+
+## Incoming Requests
+
+The eWallet allows you to configure different aspects of the incoming requests.
+We try to provide sane default values, but if needed, the following environment variables are configurable:
```
T-Dnzt: There are still mentions of `environment variable` in this doc, we should switch all those to `settings`.

PR #597, opened 2018-12-17
Stricter Redirect URL prefixes comparison

### LINKED ISSUE #583: Possible Bypass For URL Validator
# Subject of the issue
In https://github.com/omisego/ewallet/blob/master/docs/setup/advanced/env.md, the following documentation is written for the REDIRECT_URL_PREFIXES: 

>External redirects
Some features such as email verification allows redirects to URIs external to the eWallet. For security reasons we do not allow redirects to arbitary URIs, unless the prefix is whitelisted.

>REDIRECT_URL_PREFIXES: A comma-separated list of prefixes that are allowed to be redirected to. For example, REDIRECT_URL_PREFIXES=https://example.com,pos-client://example.com allows redirects to https://example.com/some_url and pos-client://example.com/some_deep_linked_url.

By setting the url as https://example.com, there are some possible bypasses

# Your environment

https://github.com/omisego/ewallet/blob/master/apps/ewallet/lib/ewallet/web/url_validator.ex

# Steps to reproduce

I have used the code from url_validator.ex (with some hard coded urls for visibility)

```
defmodule UrlValidator do
  @moduledoc """
  This module validates a url.
  """

  @doc """
  Checks that the given url is allowed as a redirect url.
  """
  def allowed_redirect_url?(url) do
    base_url = "https://example.com"
    prefixes = ["https://google.com", "https://example2.com"]
    allowed = [base_url | prefixes]

    Enum.any?(allowed, fn prefix -> String.starts_with?(url, prefix) end)
  end
end

IO.puts UrlValidator.allowed_redirect_url?("https://example.com.notexample.com")
IO.puts UrlValidator.allowed_redirect_url?("https://example.com@notexample.com")
IO.puts UrlValidator.allowed_redirect_url?("https://example.comnotexample.com")
```

This produces the following output
```
$elixir main.exs
true
true
true
```

# Expected behavior

None of the 3 inputs are of the host example.com. Thus, it should return false.

# Actual behavior

True is returned.

# Mitigation suggestions
1) Make sure to state in the documentation that you need to end the whitelisted REDIRECT_URL_PREFIXES with a "/" character.

2) Instead of checking the whole path, check the host domain name against a whitelist.

unnawut:
Issue/Task Number: #583
Closes #583 

# Overview

This PR imposes a stricter redirect URL prefixes to prevent bypassing the checks with a domain name that contains the whole prefix.

# Changes

- Updated `EWallet.Web.UrlValidator.allowed_redirect_url?/1` to add trailing slashes to both the given url and the allowed prefixes before the comparison.
- Add more tests with values as suggested by @jpopxfile

# Implementation Details

- `String.start_with/2` is used instead of regex as it should be more performant.
- Also the trailing slashes are added during the comparison instead of when the values are saved to the database. This should be safer as this makes the check independent of the functions that save the settings to the database.

# Usage

Set `redirect_url_prefixes` settings to "http://example.com". Then try reset password with `redirect_url: http://example.com.phishing.com/redirect_url?email={email}&token={token}`. The server should return a "redirect_url is not allowed" error.

# Impact

No changes to the DB schema or API specs

### DIFF
--- a/apps/ewallet/lib/ewallet/web/url_validator.ex
+++ b/apps/ewallet/lib/ewallet/web/url_validator.ex
@@ -4,7 +4,7 @@ defmodule EWallet.Web.UrlValidator do
   """
 
   @doc """
-  Checks that the given url is allowed as a redirect url.
+  Checks that the given url is allowed as a redirect url in the application settings.
   """
   def allowed_redirect_url?(url) do
     base_url = Application.get_env(:ewallet, :base_url)
@@ -14,6 +14,9 @@ defmodule EWallet.Web.UrlValidator do
     Enum.any?(allowed, fn prefix -> allowed_redirect_url?(url, prefix) end)
   end
 
+  @doc """
+  Checks that the given url is allowed by the given prefix
+  """
   def allowed_redirect_url?(url, prefix) do
     # Add trailing slashes to prevent urls such as 'https://example.comnotexample.com'
     # matching 'https://example.com'
--- a/apps/ewallet/lib/ewallet/web/url_validator.ex
+++ b/apps/ewallet/lib/ewallet/web/url_validator.ex
@@ -11,6 +11,19 @@ defmodule EWallet.Web.UrlValidator do
     prefixes = Application.get_env(:ewallet, :redirect_url_prefixes)
     allowed = [base_url | prefixes]
 
-    Enum.any?(allowed, fn prefix -> String.starts_with?(url, prefix) end)
+    Enum.any?(allowed, fn prefix -> allowed_redirect_url?(url, prefix) end)
+  end
+
+  def allowed_redirect_url?(url, prefix) do
+    # Add trailing slashes to prevent urls such as 'https://example.comnotexample.com'
+    # matching 'https://example.com'
+    url = trailing_slashed(url)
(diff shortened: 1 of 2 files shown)


apps/ewallet/lib/ewallet/web/url_validator.ex
```diff
@@ -4,13 +4,29 @@ defmodule EWallet.Web.UrlValidator do
...
+  end
+
+  @doc """
+  Checks that the given url is allowed by the given prefix
+  """
+  def allowed_redirect_url?(url, prefix) do
+    # Add trailing slashes to prevent urls such as 'https://example.comnotexample.com'
+    # matching 'https://example.com'
+    url = trailing_slashed(url)
+    prefix = trailing_slashed(prefix)
```
sirn: Use `URI.merge`:

```
iex(5)> "https://foo.bar" |> URI.merge("/foo/bar/baz") |> URI.to_string()
"https://foo.bar/foo/bar/baz"
```

unnawut: Since `URI.merge("https://foo.bar/some/segment", "/")` would return `"https://foo.bar/"` losing the URI segments, I've reverted back to basic string matching instead.

PR #603, opened 2018-12-19
Add "unsigned_integer" setting type

jarindr:
Issue/Task Number: #580 

# Overview
Improve validation for integer type in setting, otherwise it will accept minus number.
# Changes

- Improve integer validation.
- Remove warning problem from deprecated function.
- Add button disable case.

# Usage
Calling setting Api with negative integer will return error.


apps/ewallet_config/lib/ewallet_config/setting_validator.ex
```diff
@@ -4,6 +4,32 @@ defmodule EWalletConfig.SettingValidator do
...
+    do_validate_positive_integer(changeset)
+  end
+
+  def validate_positive_integer(changeset, %{type: "integer"}) do
+    do_validate_positive_integer(changeset)
+  end
+
+  def validate_positive_integer(changeset, _), do: changeset
+
+  def do_validate_positive_integer(changeset) do
```
mederic-p: Can be private (`defp`)

mederic-p: Also, can we add a test on this?

PR #606, opened 2018-12-20
Activity Logs Endpoints

### LINKED ISSUE #574: Add endpoint to retrieve activity logs

mederic-p:
Issue/Task Number: 574
Closes #574 

# Overview

This PR adds the `api/admin/activity_log.all` endpoint to get a paginated list of activity logs.

# Changes

- Add new endpoint and all related files
- Add a new gate to preload and format the `target` and `originator`
- Add `originator_identifier` and `target_identifier` to `activity_log` table

# Implementation Details

The `activity_log` resource is a little bit different from other resources that we currently have as it doesn't contains direct relations to its `originator` and `target`. In fact only the `uuid` and `type` were stored until now.

In order for frontend to display relevant informations about an activity log, it's necessary when possible to have the full representation of the originator and target.
In this PR, we first query the paginated results with the given params in order to get a list of activity logs, then we group all targets and originators by type and perform one query for each schema to get the objects. Finally, we inject the originator and target for each log in the original list and we render it to the user.

The `target_identifier` and `originator_identifier` are also added to `activity_log` table to simplify the filtering process.

These identifier will most of the time be the `id` of the resource but can be anything, ie: for wallets it's the `address`. There is a mapping of which field to use for each resource.
This endpoint does NOT currently support multi level filtering, so only the keys at the root of the json can be used to filter/sort.

# Usage

```
curl -X POST "http://some_url/api/admin/activity_log.all" \
-H "accept: application/vnd.omisego.v1+json" \
-H "Authorization: OMGAdmin some_auth" \
-H "Content-Type: application/vnd.omisego.v1+json" \
-d "{\"page\":1,\"per_page\":10,\"match_all\":[],\"match_any\":[],\"sort_by\":\"created_at\",\"sort_dir\":\"desc\"}"
```

# Impact

- New migrations
- Existing records will not have a `target_identifier` or `originator_identifier` set.
- New API


apps/ewallet/lib/ewallet/gates/activity_log_gate.ex
```diff
@@ -0,0 +1,138 @@
...
+  end
+
+  # Extract and prepare data that need preloading in the following format
+  # %{
+  #   Entity1: [uuid1, uuid2, uuid3],
+  #   Entity2: [uuid1, uuid2],
+  #   ....
+  # }
+  defp format_preload_values(data, module_mapper) do
+    Enum.reduce(data, %{}, fn activity_log, acc ->
```
unnawut: I think Elixir's immutability might make this code quite intensive to build from top down. So it might be better to use more native Enum functions + avoid nested loops altogether.

Maybe something like this instead (untested)?

```ex
defp format_preload_values(data, module_mapper) do
  {data, []}
  |> extract_entity_uuids(:originator_type, :originator_uuid, module_mapper)
  |> extract_entity_uuids(:target_type, :target_uuid, module_mapper)
  |> elem(1)
  |> Enum.uniq()
  |> Enum.group_by(&elem(&1, 0), &elem(&1, 0))
end

defp extract_entity_uuids({raw, extracted}, schema_type, record_identifier, module_mapper) do
  extracted = 
    Enum.reduce(raw, extracted, fn activity_log, acc ->
      schema = activity_log |> Map.get(schema_type) |> ActivityLog.get_schema()

      if preloadable?(schema, module_mapper) do
        [{schema, Map.get(activity_log, record_identifier} | acc]
      else
        acc
      end
    end

  {raw, extracted}
end

#
# Delete code below
#
#
# defp format_preload_values(data, module_mapper) do
#   Enum.reduce(data, %{}, fn activity_log, acc ->
#     acc
#     |> add_preload_if_valid(
#       ActivityLog.get_schema(activity_log.originator_type),
#       activity_log.originator_uuid,
#       module_mapper
#     )
#     |> add_preload_if_valid(
#       ActivityLog.get_schema(activity_log.target_type),
#       activity_log.target_uuid,
#       module_mapper
#     )
#   end)
# end
#
# # Checks if the given schema needs to be loaded.
# # If no -> do nothing
# # If yes -> adds an array of uuids for the given schema key
# defp add_preload_if_valid(preloads, schema, uuid, module_mapper) do
#  case preloadable?(schema, module_mapper) do
#    true ->
#      preloads
#      |> Map.put_new(schema, [])
#      |> add_uuid_if_not_present(schema, uuid)

#     false ->
#      preloads
#  end
# end

# # Adds the given uuid to the list for the schema if not already present
# defp add_uuid_if_not_present(preloads, schema, uuid) do
#  uuids =
#    case !Enum.member?(preloads[schema], uuid) do
#      true -> [uuid | preloads[schema]]
#      false -> preloads[schema]
#    end

#   Map.put(preloads, schema, uuids)
# end
```

apps/ewallet/lib/ewallet/web/v1/ModuleMapper.ex
```diff
@@ -0,0 +1,80 @@
+defmodule EWallet.Web.V1.ModuleMapper do
```
unnawut: filename: `module_mapper.ex`

apps/ewallet/test/ewallet/gates/activity_log_gate_test.exs
```diff
@@ -0,0 +1,47 @@
...
+      activity_log2 =
+        insert(:activity_log, %{
+          originator_type: ActivityLog.get_type(User),
+          originator_uuid: user.uuid
+          # target is set to system in the factory
+        })
+
+      activity_logs = [activity_log1, activity_log2]
+
+      assert_raise KeyError, fn -> activity_log1.originator end
```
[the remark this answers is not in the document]
unnawut: Closing since this is for testing.

unnawut:
Looks good to merge to me so we can lock down features for `v1.1`.

PR #650, opened 2019-01-14
Add Dialyzer in build steps

sirn:
Issue/Task Number: #663

# Overview

Run Dialyzer as part of the build process in order to ensure code are properly written.

# Changes

- Cleanup build steps to use a shared workspace instead of cache
- Add Dialyzer step

# Impact

You will probably see `dialyze` step failing in all PRs until everything is fixed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,24 @@ jobs:
             make check-credo
           environment:
             MIX_ENV: test
+
+  dialyze:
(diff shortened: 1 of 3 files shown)


sirn:
FWIW `dialyze` is currently not required to be passed (unlike `test` and `lint`). It will be required as soon as we managed to fix all dialyzer errors.

PR #663, opened 2019-01-16
Better handling for export failures

### LINKED ISSUE #657: Prevent download of generated exports when adapter is invalid
- Prevent download of generated exports when adapter is invalid
- Ensure exports fail after a certain time (when the GenServer dies?)

T-Dnzt:
Issue/Task Number: 657
closes #657 

# Overview

This PR improves a few things on the export flow.

# Changes

- Set an export as failed if the process terminates
- Properly handle upload failures to GCS / AWS
- Only return exports that have been generated with the current adapter
- Prevent the `get` for exports that have been generated with another adapter


apps/admin_api/test/admin_api/v1/controllers/admin_auth/export_controller_test.exs
```diff
@@ -67,6 +67,19 @@ defmodule AdminAPI.V1.ExportControllerTest do
...
+      user = get_test_admin()
+      insert(:export, adapter: "local", user_uuid: user.uuid)
+      insert(:export, adapter: "local", user_uuid: user.uuid)
+      insert(:export, adapter: "gcs", user_uuid: user.uuid)
+
+      response = admin_user_request("/export.all", %{})
+      exports = response["data"]["data"]
+
+      assert response["success"]
+      assert length(exports) == 2
```
unnawut: Can you assert the two records for `adapter: "local"` too? Just in case the results are a side effect from other tests.

apps/ewallet/lib/ewallet/exporters/csv_exporter.ex
```diff
@@ -61,11 +61,32 @@ defmodule EWallet.CSVExporter do
      }}
   end
 
+  def terminate(reason, %{export: %{status: "processing"} = export}) do
+    {:ok, _} =
+      AdapterHelper.store_error(export, "The server crashed during processing.", inspect(reason))
```
unnawut: Minor nitpick. Can we say "The **export** server crashed ..."? Because it could mean the web server as well.

Or "The **exporter** crashed..."

apps/ewallet/lib/ewallet/exporters/csv_exporter.ex
```diff
@@ -61,11 +61,32 @@ defmodule EWallet.CSVExporter do
...
   def handle_cast(:upload, state) do
     adapter_module = get_adapter_module()
-    adapter_module.upload(state)
 
-    {:stop, :normal, state}
+    case adapter_module.upload(state) do
+      {:ok, export} ->
+        {:stop, :normal, %{state | export: export}}
+
+      {:error, %{status: "failed"} = export} ->
```
unnawut: Can you add a comment here to document why `{:error, %{status: "failed"}}` acts the same as `{:ok, export}` (i.e. does not call `AdapterHelper.store_error/2` like below)?

PR #672, opened 2019-01-17
Fix E2E configuration string splitting

sirn:
Issue/Task Number: #667 

# Overview

When parsing long string to `base64` UNIX command it will (unfortunately) helpfully split it into multiple lines to help with readability. Sadly enough, we're passing this string will newline to the `ReleaseTasks.config_base64` which make long string undecode-able on Elixir side.

# Changes

- Strip newline after converting base64 on shell side

# Impact

I can go to gym. Bye.


sirn:
@mederic-p can you push fixes to E2E directly to this branch (if you found the cause)? Thanks.

.circleci/config.yml
```diff
@@ -393,6 +398,7 @@ workflows:
             branches:
               ignore:
                 - master
+                - 667-everything-is-so-broken
```
unnawut: This still needed for the merge?

.circleci/config.yml
```diff
@@ -403,6 +409,7 @@ workflows:
             branches:
               only:
                 - master
+                - 667-everything-is-so-broken
```
unnawut: This still needed for the merge?

sirn: Should be removed before merge; but I want to get E2E working before removing this.

apps/ewallet_config/mix.exs
```diff
@@ -32,7 +32,8 @@ defmodule EWalletConfig.MixProject do
   def application do
     [
       extra_applications: [:logger],
-      mod: {EWalletConfig.Application, []}
+      included_applications: [:goth],
```
unnawut: 👍

sirn:
It’s happening!!!1

PR #738, opened 2019-01-28
Add reset frequency to cached balance

### LINKED ISSUE #367: Add reset frequency to cached balance
Currently, cached balances are either calculated from scratch (strategy `since_beginning`) or from the last caching point (strategy `since_last_cached`). 

The `since_last_cached` strategy should feature an additional option (`reset_frequency`) to allow a provider to specify a once in a while reset of the cache. Basically, every X times (as specified in `reset_frequency`, the `since_last_cached` strategy will switch to a `since_beginning` strategy. 

This feature prevents an invalid cached balance from running rampant forever, since it will be reset and re-calculated from scratch. Note that incorrect balances __should not happen__, but this measure will ensure that if it does happen once, the problem will solve itself and give us time to figure out the problem without breaking the whole system.

mederic-p:
Issue/Task Number: 367
Closes #367 

# Overview

This PR adds a new setting `balance_caching_reset_frequency` which allows the caching balance mechanism to be forced to re-calculate the balances from the beginning every `x` times.
Note that this setting will work only when the caching strategy is set to `since_last_cached` which is not the default.

# Changes

- Add `balance_caching_reset_frequency` setting (default to 10)
- Update admin panel configuration page to allow admins to update this setting

# Implementation Details

Added a new counter on cached_balance that is incremented by one every time the caching is processed. When this counter reaches the `reset frequency`, the balances are re-calculated since the beginning and the counter is reset to 1.
If the frequency is set to `0` or `nil` and the strategy is set to `since_last_cached`, it'll keep caching from the previously cached balances.

# Usage

Set the caching strategy to `since_last_cached` and the frequency to some value (ie: `5`) and run `LocalLedger.CachedBalance.cache_call()` 5 times.

# Impact

- This PR requires to run migrations on the local_ledger_db
- This PR requires the settings to be updated!


apps/local_ledger/test/local_ledger/cached_balance_test.exs
```diff
@@ -138,11 +162,124 @@ defmodule LocalLedger.CachedBalanceTest do
              }
     end
 
+    test "increase the cached_count when calculating when strategy = 'since_last_cached'", %{
```
unnawut: Can you recheck the test names' wordings. Seem to have two "when"s for each.

docs/setup/advanced/settings.md
```diff
@@ -37,6 +37,7 @@ Some features such as email verification allows redirects to URIs external to th
 The local ledger offers a caching mechanism for wallets in order to boost the calculation speed (in case you have millions of transactions).
 
 - `balance_caching_strategy`: Specify if new cached wallets should be computed using a previous cache or by recalculating everything from scratch.
+- `balance_caching_reset_frequency`: Specify after how many times a cached balance needs to be recalculated from the beginning and not use the last cached balance.
```
unnawut:
```suggestion
- `balance_caching_reset_frequency`: Specify that a cached balance needs to be recalculated at the n<sup>th</sup> time of usage.
```

docs/setup/advanced/settings.md
```diff
@@ -38,10 +38,12 @@ The local ledger offers a caching mechanism for wallets in order to boost the ca
...
 
-Strategies available:
+  Strategies available:
+  - `since_beginning`: Recalculate the balance since the beginning of time.
+  - `since_last_cached`: Use the last cached balance, adds the transactions that happened since and saves the result in a new cached balance.
 
-- `since_beginning`: Recalculate the balance since the beginning of time.
-- `since_last_cached`: Use the last cached balance, adds the transactions that happened since and saves the result in a new cached balance.
+
+- `balance_caching_reset_frequency`: Specify that a cached balance needs to be recalculated at the n<sup>th</sup> time of usage when using a `since_last_cached` caching strategy. Setting this to 0 means that cached balances will never be re-calculated from the beginning.
```
unnawut: This one leaves a room open for interpretation when `since_last_cached` is not used. Suggesting...

```suggestion
- `balance_caching_reset_frequency`: Specify that a cached balance needs to be recalculated at the n<sup>th</sup> time of usage. Setting this to 0 means that cached balances will never be re-calculated from the beginning. Applicable only when `since_last_cached` caching strategy is used.
```

Issue #732, opened 2019-01-24
Run E2E tests for version branches


sirn:
Also need to create corresponding `ewallet-v1.1` branch or something in omisego/e2e repo (for `v1.1`), and use that in E2E, so we can add more test to end-to-end without breaking older versions (especially when there's incompatible changes).

sirn:
/u/nebali is like, sirn, can you stop adding tasks to 1.1

sirn:
This is actually done so bye

PR #749, opened 2019-01-29
Update Circle CI configuration to use versioned branch for E2E

sirn:
Issue/Task Number: #732 

---

This PR changes the building process so that for an `ewallet/v1.1` branch of https://github.com/omisego/e2e repo is being used for `v1.1` instead of the `master` branch. This is to prevent future updates to E2E (e.g. blockchain integration) from breaking testing pipeline in older versions in the future.

---

This PR also updates the Docker tag convention to resembles the older convention, as in:

* `omisego/ewallet:dev` is now `omisego/ewallet:latest` (same as before)
* Tip of versioned branches (`v*`) now has corresponding `v*-dev` tag.
* All commits from versioned branches and master branches now has `<shortref>` tag.

Basically, we’re reverting breaking changes introduced in earlier build pipeline in favor for an old behavior (which were a lot less confusing).


sirn:
@unnawut @T-Dnzt need comment on Docker tags in general. Before we moved to Circle it was `:latest` then I changed master to `:dev` because Docker will pull `:latest` by default and I don’t want people to pull latest dev… but that make `:latest` really outdated (pre-Circle) because we never updated it.

PR #763, opened 2019-01-31
Fix assets path in releases

### LINKED ISSUE #756: Admin panel & Swagger UI assets could not be loaded
# Subject of the issue

Admin panel & Swagger UI assets are not being served on the latest docker image of `omisego/ewallet:latest`.

# Your environment

- MacOS 10.13.6
- Docker version 18.09.1, build 4c52b90
- docker-compose version 1.23.2, build 1110ad01
- Docker image: omisego/ewallet, latest, 21ef67e9fd36, 8 hours ago, 241MB

# Steps to reproduce

Run `docker-compose up -d` following the setup instructions in the README.md.

Then try to access http://localhost:4000/admin, http://localhost:4000/api/admin/docs.ui and http://localhost:4000/api/client/docs.ui

# Expected behavior

- http://localhost:4000/admin displays the admin panel front end successfully.
- http://localhost:4000/api/admin/docs.ui renders the Swagger UI for Admin API.
- http://localhost:4000/api/client/docs.ui renders the Swagger UI for Client API.

# Actual behavior

- http://localhost:4000/admin displays:
    > The assets are not available. If you think this is incorrect, please make sure that the front-end assets have been built.
- http://localhost:4000/api/admin/docs.ui returns the page but assets are all 404
- http://localhost:4000/api/client/docs.ui returns the page but assets are all 404

sirn:
Closes #756 

Fix unable to access /admin due to path resolving in releases. `__DIR__` and other kernel forms are expanded at compile time, so they’re always returning the wrong path. This PR fixes the issue by resolving static path using `Application.app_dir` instead.

Please also verify that `omg.server` and other things are working fine.


sirn:
Ah darn swagger still broke because assets doesn’t get included in any apps.

sirn:
Btw if you're running `mix omg.server` you probably want to move your `public/` and `private/`:

* `/public` -> `apps/url_dispatcher/priv/static/public`
* `/private` -> `apps/url_dispatcher/priv/static/private`

(Moving as-is should be fine)

🙇

mederic-p:
I'm getting these warnings when compiling:
```
warning: function EWallet.Helper.static_dir/1 is undefined or private
  lib/ewallet_config/storage/local.ex:28
```

```
warning: unused import Phoenix.Controller.redirect/2
  lib/url_dispatcher/plug.ex:18
```

unnawut:
1. `mix omg.server` is working well! All below works:

    * /admin page is accessible
    * Swagger works
    * Local uploads are working properly
    * Local exports are working properly

2. Endpoints are not served when running `ewallet foreground`:

    ```shell
    _build/prod/rel/ewallet/bin/ewallet foreground
    12:25:49.623 [info] AppSignal disabled.
    12:25:49.816 [info] UrlDispatcher.Plug disabled.
    ```

    But it can be started by manually passing `SERVE_ENDPOINTS`:

    ```shell
    SERVE_ENDPOINTS=true _build/prod/rel/ewallet/bin/ewallet foreground
    12:27:23.057 [info] AppSignal disabled.
    12:27:23.301 [info] Running UrlDispatcher.Plug with Cowboy on port 4000.
    12:27:23.301 [info] Running EWalletAPI.V1.Endpoint WebSocket endpoint at /api/client/socket.
    12:27:23.301 [info] Running AdminAPI.V1.Endpoint WebSocket endpoint at /api/admin/socket.
    ```

    Now when the endpoints are serving, these work!

    * /admin page is accessible
    * Swagger works
    * Local uploads are working properly
    * Local exports are working properly

3. I'm seeing the same warnings as @mederic-p as well.

sirn:
SERVE_ENDPOINTS is intentional. In rootfs/entrypoint we default this to true if user run it via alias (e.g. docker run omisego/ewallet:latest foreground) rather than via bin/ewallet.

I’m not sure if there’s a way around this because otherwise bin/ewallet console would start the server which might not be the behavior ones want (we use that mostly for debug right?)

Otherwise thanks for double checking!

unnawut:
Right. Since the entrypoint's code isn't changed and the `SERVE_ENDPOINTS` is indeed set there. I think this is good to merge.

Will regression test again once the remaining `v1.1` PRs/issues are merged.

unnawut:
Oh also... I think `EWallet` is not a dependency in `EWalletConfig`, and intended not to be I believe.

Is it possible to move `EWallet.Helper.static_dir/1` to `Utils` app?

sirn:
Oh I didn't noticed utils app was now a thing. Will move it there.

sirn:
Moving to utils going to take a bit more time (only a bit), since it depends on `to_boolean` which happens to be in `EWallet.Helper` so I need to move that too.

sirn:
I've moved all relevant files, please review again. Thanks!

PR #777, opened 2019-02-04
Migrate BALANCE_CACHING_FREQUENCY to EWalletConfig

### LINKED ISSUE #767: Migrate BALANCE_CACHING_FREQUENCY to EWalletConfig
Right now `BALANCE_CACHING_FREQUENCY` still use env var.

unnawut:
Issue/Task Number: #767 
Closes #767

# Overview

This PR implements the migration for `BALANCE_CACHING_FREQUENCY` environment variable to the database.

# Changes

- Added the default value for `:balance_caching_frequency`
- Added `:balance_caching_frequency` to `config --migrate`
- Implemented `EWalletConfig.BalanceCachingSettingsLoader`
- Updated `EWalletConfig.SettingsLoader` to support multiple custom loaders instead of just `FileStorageSettingsLoader`
- Added `:namespace` application config to all apps that are missing them

# Implementation Details

Since the balance caching frequency is setup during application start and not monitored for change, we needed to implement a custom loader that would trigger and apply the changes to the scheduler when the settings is updated. This is done in similar manner to `FileStorageSettingsLoader`.

# Usage

Run `mix omg.config --migrate` on a machine that has the env var configured should show a message that the env var will be migrated. For example:

```shell
$ BALANCE_CACHING_FREQUENCY="* * * * *" mix omg.config --migrate

Starting the settings migration task...
The following settings will be populated into the database:

  - balance_caching_frequency: "* * 2 * *"

Are you sure to migrate these settings to the database? [Yn]
```

And of course the new change should be reflected after migrating.

# Impact

Run `mix omg.config --migrate` (on development) or `ewallet config --migrate` (on production).

### DIFF
--- a/apps/ewallet_config/lib/ewallet_config/loaders/balance_caching_settings_loader.ex
+++ b/apps/ewallet_config/lib/ewallet_config/loaders/balance_caching_settings_loader.ex
@@ -28,22 +28,31 @@ defmodule EWalletConfig.BalanceCachingSettingsLoader do
     scheduler = Application.get_env(app, :scheduler)
     frequency = Application.get_env(app, :balance_caching_frequency)
 
-    update_frequency(frequency, scheduler)
+    case alive?(scheduler) do
+      true -> update_frequency(scheduler, @job_name, frequency)
+      false -> {:error, :scheduler_not_running}
+    end
   end
 
-  defp update_frequency(nil, _scheduler) do
-    {:error, :frequency_not_found}
-  end
+  defp alive?(nil), do: false
 
-  defp update_frequency(frequency, scheduler) when is_binary(frequency) do
+  defp alive?(module) when is_atom(module), do: module |> GenServer.whereis() |> alive?()
+
+  defp alive?(pid) when is_pid(pid), do: Process.alive?(pid)
+
+  defp alive?(_), do: false
+
+  defp update_frequency(_, _, nil), do: {:error, :frequency_not_found}
+
+  defp update_frequency(scheduler, job_name, frequency) when is_binary(frequency) do
     case Parser.parse(frequency) do
-      {:ok, cron_expression} -> update_frequency(cron_expression, scheduler)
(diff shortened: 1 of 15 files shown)


apps/ewallet_config/lib/ewallet_config/loaders/balance_caching_settings_loader.ex
```diff
@@ -0,0 +1,62 @@
...
+  Maps the DB settings to the configuration needed for balance caching mechanism.
+  """
+  require Logger
+  alias Crontab.CronExpression.Parser
+
+  @job_name :cache_all_wallets
+
+  @spec load(atom()) :: :ok | {:error, :scheduler_config_not_found}
+  def load(app) do
+    namespace = Application.get_env(app, :namespace)
```
sirn: Wouldn't it be better if we just define `scheduler:` instead, so we can be a bit more flexible about scheduler?

apps/ewallet_config/lib/ewallet_config/loaders/balance_caching_settings_loader.ex
```diff
@@ -0,0 +1,62 @@
...
+    config = Module.concat(namespace, Config)
+    scheduler = Module.concat(namespace, Scheduler)
+
+    :ok = scheduler.delete_job(@job_name)
+
+    case config.read_scheduler_config() do
```
sirn: Can we not just leverages the existing `{:apply, MFA}` in `jobs:` key?

unnawut:
Removing upgrade path flag as this PR will go with `v1.1.0` which already has config migration as the upgrade path.

apps/ewallet_config/lib/ewallet_config/loaders/balance_caching_settings_loader.ex
unnawut: Much simpler now. Thanks!!

PR #788, opened 2019-02-07
Fix Docker push

### LINKED ISSUE #787: Make version detection in Circle more strict
Right now a branch named `version-bump-1.1.0` will match the `v*` rules and get pushed to Docker as `ersion-bump-1.1.0-dev` (removing the first `v` and add `-dev`) which… is… probably… something… we… don't… want.

sirn:
Closes #787 

* Fixes stable tag detection
* Fixes tag detection for a version branch
* Fixes a missing ssh-keyscan during ls-remote

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -350,6 +350,9 @@ jobs:
           name: Publishing Docker image
           command: |
             set -xe
+
+            mkdir -p ~/.ssh
+            ssh-keyscan github.com >> ~/.ssh/known_hosts
             docker load -i ~/caches/docker-layers.tar
 
             IMAGE_TAG=""
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -371,11 +371,13 @@ jobs:
                     esac
                 done
 
-                # In case the commit is HEAD of master branch, also tag stable.
+                # In case the commit is HEAD of latest version branch, also tag stable.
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -371,11 +374,13 @@ jobs:
...
 
-                # In case the commit is HEAD of master branch, also tag stable.
+                # In case the commit is HEAD of latest version branch, also tag stable.
                 if [ -n "$CIRCLE_REPOSITORY_URL" ] && [ -n "$CIRCLE_SHA1" ]; then
-                    _ref="refs/heads/master"
-                    _head="$(git ls-remote "$CIRCLE_REPOSITORY_URL" "$_ref" | awk '{ print $1 }')"
-                    if [ "$CIRCLE_SHA1" = "$_head" ]; then
+                    _stable_head="$(
+                        git ls-remote --heads "$CIRCLE_REPOSITORY_URL" "v*" |
+                        awk '/refs\/heads\/v[0-9]+\.[0-9]+$/ { LH=$1 } END { print LH }')"
```
unnawut: Won't these two lines mean that branch commits (not released/tagged) will also be considered stable?

sirn: These `stable` check only run when building against tags so it shouldn't happen.

PR #793, opened 2019-02-07
Always restart Goth on settings load

sirn:
Fix an issue where Goth fail to load credentials on application start due to Goth app has already been started and env never got populated there.


sirn:
Not sure if it will actually fix the issue, but this is the only one I can think about.

PR #806, opened 2019-02-14
ExternalLedgerDB.Wallet schema

### LINKED ISSUE #690: Blockchain wallet schema
Depends on #689 

In this first implementation, the private key for the wallet can be encrypted and stored in the DB with the encryption key being stored in an environment variable. The security around the storage of the encryption key will be added in another issue.

The schema will at least need the following fields:

- `uuid`
- `id`
- `type`: `hot` / `cold`
- `encrypted_private_key`
- `public_key`
- `address`
- `primary`
- `timestamps`

A set of rules should also be available for this. Like balance shouldn't go above `xxx`, with a few options like notifying or transferring to a cold wallet.

unnawut:
Issue/Task Number: #690 
Closes #690
Requires #807 

# Overview

This PR adds a new `ExternalLedgerDB` app to manage schemas and database rules for all external ledgers that we may support. Currently namely Ethereum network and OMG plasma network.

# Changes

- Upgraded `:ecto_sql` to fix Ecto ignoring `:queue_target` and `:queue_interval` options
- Added a new `ExternalLedgerDB` sub-app
- Added `ExternalLedgerDB.Wallet` and its related migration and tests
- Added `Utils.Ledgers.WalletSchema` behaviour and enforce on `ExternalLedgerDB.Wallet` and `LocalLedgerDB.Wallet`

# Implementation Details

This is a straight-forward addition of a new sub-app and `ExternalLedgerDB.Wallet` schema.

A `Utils.Ledgers.WalletSchema` was added and enforced on `ExternalLedgerDB.Wallet` and `LocalLedgerDB.Wallet` with anticipation that the `EWallet` and its related apps will be interacting with local & external wallets interchangeably. With this behavioural enforcement we should be able to do the routing between the two ledgers much easier than without the enforcement.

# Usage

Not yet. DB schema and migrations only. No API to interact with.

# Impact

New DB migration. No API specs changed.


apps/external_ledger_db/lib/external_ledger_db/wallet.ex
```diff
@@ -0,0 +1,174 @@
...
+  @cold "cold"
+
+  def hot, do: @hot
+  def cold, do: @cold
+
+  @ethereum "ethereum"
+  @omg_network "omg_network"
+
+  def ethereum, do: @ethereum
+  def omg_network, do: @omg_network
```
T-Dnzt: This shouldn't be hardcoded. The adapters should have a way to register themselves.

apps/external_ledger_db/lib/external_ledger_db/wallet.ex
```diff
@@ -0,0 +1,174 @@
...
+    |> validate_required([
+      :address,
+      :adapter,
+      :primary,
+      :type,
+      :public_key,
+      :metadata,
+      :encrypted_metadata
+    ])
+    |> validate_inclusion(:adapter, [@ethereum, @omg_network])
```
T-Dnzt: Ethereum and the OMG network are not exclusive, they work together. The same wallet (same Ethereum address) can have funds on Ethereum AND Plasma at the same time. This should validate the inclusion within the registered list of adapters instead. We'll later have an adapter that will basically be `ethereum + omg_network` (but we could also have an adapter just for `ethereum`, which will be the case at first).

apps/external_ledger_db/lib/external_ledger_db/wallet.ex
```diff
@@ -0,0 +1,174 @@
...
+  @doc """
+  Validate the wallet attributes.
+  """
+  @spec changeset(%__MODULE__{}, map()) :: Ecto.Changeset.t()
+  def changeset(%__MODULE__{} = wallet, attrs) do
+    wallet
+    |> cast(attrs, [
+      :address,
+      :adapter,
+      :primary,
```
T-Dnzt: Should we have a specific changeset for the primary hot wallet? Since it won't happen often. An alternative would be to store it in the settings instead of inside the DB. Thoughts?

apps/utils/lib/ledgers/wallet_schema.ex
```diff
@@ -0,0 +1,27 @@
...
+  Defines common behaviours for ledger wallet schemas.
+  """
+
+  @callback all([String.t()]) :: [Ecto.Schema.t()]
+  @callback get(String.t()) :: Ecto.Schema.t() | nil
+  @callback get_by(Keyword.t() | map()) :: Ecto.Schema.t() | nil
+  @callback get_or_insert(map()) :: {:ok, Ecto.Schema.t()} | {:error, Ecto.Changeset.t()}
+  @callback insert(map()) :: {:ok, Ecto.Schema.t()} | {:error, Ecto.Changeset.t()}
+  @callback touch([String.t()]) :: {integer(), nil | [term()]}
+  @callback lock([String.t()]) :: [Ecto.Schema.t()]
```
T-Dnzt: Not sure we need this, local wallets and ledger wallets are not supposed to work in the same way. You won't be able to lock a blockchain wallet (the balances are kept on the blockchain anyway, so the blockchain is responsible for preventing double spends). 

When making a transaction, a few scenarios can occur: 

- local to local (like now)
- local to blockchain (in the background, the primary hot wallet will be used to send the funds on the blockchain, but the balance will be deducted from the local wallet)
- blockchain to local (the funds will be received in the hot wallet and will credit the appropriate internal wallet, default to the master account primary wallet)

Reminder of the pooled model:
Local Wallet A ----> Funds pooled into Blockchain Wallet X
Local Wallet B ---/
Local Wallet C--/

apps/external_ledger_db/lib/external_ledger_db/wallet.ex
```diff
@@ -0,0 +1,174 @@
...
+  def ethereum, do: @ethereum
+  def omg_network, do: @omg_network
+
+  schema "wallet" do
+    field(:address, :string)
+    field(:adapter, :string)
+    field(:primary, :boolean, default: false)
+    field(:type, :string)
+    field(:public_key, :string)
+    field(:encrypted_private_key, ExternalLedgerDB.Encrypted.String)
```
sirn: We won’t be storing private keys in the ledger DB. They’ll be managed exclusively by the `BlockchainEth` possibly as `BlockchainEthDB`, where it will store just `{address, private_key}` pair encrypted.

This is to limit the surface that have access to private keys and allow other key management scenarios in the future (say, store the key in HashiCorp Vault, using Geth unlocked where we don’t manage keys at all, etc.). See #807

apps/external_ledger_db/lib/external_ledger_db/wallet.ex
```diff
@@ -0,0 +1,174 @@
...
+
+  def ethereum, do: @ethereum
+  def omg_network, do: @omg_network
+
+  schema "wallet" do
+    field(:address, :string)
+    field(:adapter, :string)
+    field(:primary, :boolean, default: false)
+    field(:type, :string)
+    field(:public_key, :string)
```
sirn: Just some note that I’m not sure what is better, between storing keys as hex-encoded (decode on key-related operations) or store as bytes (encode on display).

sirn: Also probably need `address` as well. Ethereum addresses for instance is _not_ public key, but the last 20 bytes of a keccak256 hash of the public key. Eventually we will need to query by address, so I think it makes sense to store address as well. (This will be returned from `Blockchain.Wallet` as `wallet_id`)

unnawut: IMHO, storing hex-encoded public key is easier for me in terms of serving external requests as I don't need to deal with conversions for every requests. Also easier for dumping, exports, etc. I don't think we'll use the public key from here as bytes anywhere anyway?

Also, `address` is the first in this field declarations list, just in case you meant something else.

apps/external_ledger_db/lib/external_ledger_db/wallet.ex
unnawut: Hmm I like the idea of having it as a settings instead. 👍

apps/external_ledger_db/lib/external_ledger_db/wallet.ex
unnawut: A bit confused. I was thinking that each `ExternalLedgerDB.Wallet` should belong to one adapter…

And the same wallet with same address on ethereum and omg_plasma would be considered two `ExternalLedgerDB.Wallet`, because technically they share nothing apart from the same pub/priv keys? (which may not be true too for other chains). The system doesn’t need to know that they are related?

Also… do you mean a wallet.adapter should be an array like `wallet.adapter == ["ethereum", "omg_network"]`? If so I feel there’ll be a lot of conditioning based on that array compared to 1 wallet 1 adapter.

apps/external_ledger_db/lib/external_ledger_db/wallet.ex
unnawut: Will go with hex-encoded on a string column as this field will be used mostly for displaying & user-querying.

apps/external_ledger_db/lib/external_ledger_db/wallet.ex
T-Dnzt: Alright, so it'll be like the master account, let's do it 👍

unnawut:
Closing this PR as an `ExternalLedgerDB` is probably not needed for now.

unnawut:
Will submit a new PR for blockchain wallet handling.

Issue #689, opened 2019-01-22
Ethereum account generation in eWallet


T-Dnzt:
This issue consists in implementing a module that will generate a new Ethereum account and all the needed helper functions. From it, we should be able to get a private key, a public key and an address. This will later be used for hot wallet generation.

unnawut:
Question: Does it generate a private key directly or should it be a mnemonic phrase?

T-Dnzt:
But can't we just re-generate new private / public when needed? I don't see the usefulness of using a mnemonic phrase.

unnawut:
Ok so referring to #690 that private keys will be encrypted & stored in the database, with the single encryption key stored in environment variable. This should solve the trouble having to manage a lot of private keys.

T-Dnzt:
A library we might want to explore: https://github.com/izelnakri/eth

unnawut:
Looks like https://github.com/izelnakri/eth  is mainly a wrapper over https://github.com/mana-ethereum/ethereumex that casts and maps values. Very handy!

But given that it's been inactive since May 2018, we might want to roll our own that uses https://github.com/mana-ethereum/ethereumex directly.

unnawut:
Closed by #807 and #863

PR #807, opened 2019-02-15
Initial blockchain integration

sirn:
Issue/Task Number: #689 

# Overview

This pull request create a basis framework for integrating with Blockchain networks using pluggable adapters with a basic function for generating ECDH key (using secp256k1 curve) for Ethereum. 

# Changes

- Created `Blockchain` app and modules
- Created `BlockchainEth` module

# Implementation Details

The implementation is composed mainly of:

- `Blockchain.Backend` module — managing backend adapters
- `Blockchain.Wallet` module — client-facing API for working with wallet
- `BlockchainEth` module — backend adapter for connecting to Ethereum

The backend adapter is started by the `Blockchain.Backend`'s dynamic supervisor. These backends are intended to be spawn as one process per wallet in order to limit the knowledge of wallet private key to the smallest surface possible. 

All calls to backends must be done via the `Blockchain.Wallet` module using a unified interface, such as `generate_wallet` which will return wallet ID and public key (the ID is usually the address (in Ethereum case, this is KECCAK-256 hash of public key).

One of the goals of the implementation design is to allow the eWallet Blockchain Integration to connect to e.g. Geth node in the future, in which keys are stored by Geth and must be manually unlocked. (However, this _won't_ be implemented initially, but consideration are put to make it possible to do so in the future.)

# Usage

```
iex> Blockchain.Wallet.generate_wallet(:eth)
```

Note that wallet is currently not persisted. Key persistence is coming next!

# Impact

Right now, no.


apps/blockchain/mix.exs
```diff
@@ -0,0 +1,31 @@
...
+
+  def project do
+    [
+      app: :blockchain,
+      version: "0.1.0",
+      build_path: "../../_build",
+      config_path: "../../config/config.exs",
+      deps_path: "../../deps",
+      lockfile: "../../mix.lock",
+      elixir: "~> 1.6",
```
unnawut:
```suggestion
      elixir: "~> 1.8",
```

apps/blockchain/mix.exs
```diff
@@ -0,0 +1,31 @@
+defmodule Blockchain.MixProject do
+  use Mix.Project
+
+  def project do
+    [
+      app: :blockchain,
+      version: "0.1.0",
```
unnawut:
```suggestion
      version: "1.2.0-dev",
```

apps/blockchain/test/blockchain/backend_test.exs
```diff
@@ -0,0 +1,134 @@
...
+    def handle_call(:generate_wallet, _from, reg) do
+      {:reply, {:ok, "wallet_id", "public_key"}, reg}
+    end
+  end
+
+  setup do
+    mock_id = UUID.generate()
+    mock_key = UUID.generate()
+
+    Code.eval_string("""
```
unnawut: Curious question: Why it needs to be wrapped around `Code.eval_string`?

sirn:
I think private key storage is part of #807 because private key generation would be pretty useless if we don’t store it somewhere 🤔

apps/blockchain/lib/blockchain/backend.ex
```diff
@@ -0,0 +1,241 @@
...
+  end
+
+  defp ensure_backend_started({backend, _id} = backend_spec, {supervisor, h, registry} = state) do
+    case find_backend(backend_spec, state) do
+      {:ok, pid} ->
+        {:ok, pid, state}
+
+      {:not_started, mfargs} ->
+        retval =
+          DynamicSupervisor.start_child(supervisor, %{
```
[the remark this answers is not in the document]
sirn: Yes we would have `n * x` servers running. Erlang VM has an overhead of 309 words per process (= 618 bytes) so even if we have 10,000 wallet active at a time, overhead would be only 6MB or so (and once we have that much users, we would need a larger server anyway).

Though I think it might worth to have some kind of process pruning in the future to remove inactive wallet

T-Dnzt:
@sirn does that mean you're planning to update this PR?

sirn:
I’ll create a separate PR for it. Single issue doesn’t necessary need have only one PR, and I’d like to multiple small PRs than a single large one 🤔

apps/blockchain/test/blockchain/backend_test.exs
sirn: Because we’re dynamically returning the value of `UUID.generate()` above (which also need to be passed as state for comparison). 

Unfortunately it seems like the variable above are in function scope and cannot be accessed inside `defmodule`. AFAIK Elixir also use the same pattern quite a lot in their test case so it should be safe to use.

apps/blockchain/lib/blockchain/backend.ex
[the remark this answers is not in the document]
InoMurko: Is there a reason you would need to keep them running? Why not stop them or put them in hibernation when inactive?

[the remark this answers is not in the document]
InoMurko: Also, be aware that there's a system limit http://erlang.org/doc/man/erlang.html#system_info_process_limit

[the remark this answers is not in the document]
sirn: I’m assuming the usage of the blockchain application in eWallet will be “bursted,” as in lots of operations on a single wallet over a short period of time, so I want to keep it running for those operations. Though maybe we should do hibernation instead of outright terminating it as you mentioned after a short period of time, what do you think?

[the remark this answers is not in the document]
InoMurko: Just so you know, I don't have much knowledge of the behavior and assumptions you've made for this feature. 
One thing is that erlang can spawn processes very, very quickly and sometimes, it's just not worth of keeping them alive. As said, I don't know the full behavior. 
> [...] 2.90GHz Intel Core i7 dual core with 8GB memory can spawn 20,000 processes of an average of 3.0 μs/process of CPU time and 3.4 μs of elapsed (wall-clock) time.

But if you do keep them alive, you're competing against `erlang:system_info(process_limit)`. Thats the maximum allowed number of processes. If you reach that and don't start killing unnecessary processes, your VM will likely crash soon or start misbehaving to a point of being unusable. You can increase the proces_limit if you set the +P flag on the VM.

There's one more thing you're competing with. There's only so much that a supervisor can handle. I would say that if you expect the supervisor to handle more then 10k processes... you should start new supervisors and attaching new children there. Otherwise, if processes start crashing and the supervisor is supposed to 1. restart them and 2. start new incoming ones that are not running yet, you'll increase latency there and block execution.

I'll now share how I approach dynamic children, when I assume that when a customer *is active*, he'll remain active for a period of time:
I expand the workers state with a property called `last_activity` and `stay_alive` which I set in `init` to
```
# seconds since 0AD
 defp sec(), do: :calendar.datetime_to_gregorian_seconds(:erlang.localtime())
```
in `init` I would also set the above mentioned configurable timer:
```stay_alive = Application.fetch_env!(:wallet, __MODULE__)[:stay_alive]```
Every message that gets sent to the worker would reset the `last_activity` property in the state. When the timer gets invoked a message is sent to `handle_info` and handled this way:
```
def handle_info(:timeout, %__MODULE__{stay_alive: stay_alive} = state) do
    now = sec()
    # to miliseconds
    diff = (now - state.last_activity) * 1000

    case diff do
      diff when diff > stay_alive ->
        _ = debug_log("entity process is shutting down with state ~p", [state])
        {:stop, :shutdown, state}

      _ ->
        Process.send_after(self(), :timeout, stay_alive)
        {:noreply, state}
    end
  end
```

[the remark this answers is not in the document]
Comment (sirn) [reply apps/blockchain/lib/blockchain/backend.ex]: @InoMurko Thanks! I wholeheartedly agree with your concern.

My primary reasoning for keeping the process alive actually come from the plan to possibly allow replacing an in-DB private key storage with some sort of KMS (e.g. HashiCorp’s Vault) which may turn wallet startup to be a costly operation.

In such case, I think the best balance would be to terminate or hibernate the process after a period of inactivity. However as we don’t have a concrete plan to implement KMS soon, we might as well shutdown the process immediately after a single task as you have described, turning the wallet backend into a one-off task runner rather than a long running process.

I’ll look into it. Thanks again!

**Comment by sirn** [comment]
I’ve refactor this PR so the backend process works as a single task runner rather than long running. Please review again 🙇

File: apps/blockchain/lib/blockchain/backend.ex [hunk 5]
```diff
@@ -102,79 +100,50 @@ defmodule Blockchain.Backend do
...
   @doc """
   Handles the call call from the client API call/4.
   """
   @spec handle_call({:call, backend(), call()}, from(), state()) :: reply({:ok, any()})
   def handle_call({:call, backend_spec, func_spec}, _from, state) do
     case ensure_backend_started(backend_spec, state) do
-      {:ok, pid, state1} ->
-        {:reply, {:ok, GenServer.call(pid, func_spec)}, state1}
+      {:ok, pid} ->
+        try do
```
unnawut: Not too familiar with this code. Would it suppress exceptions from `GenServer.call(pid, func_spec)` without logging?

apps/blockchain/lib/blockchain/backend.ex
```diff
@@ -102,79 +100,50 @@ defmodule Blockchain.Backend do
...
+          {:reply, {:ok, GenServer.call(pid, func_spec)}, state}
+        after
+          GenServer.stop(pid)
```
unnawut: Another curious & dumb question. In a stop-after-executing case like this, why not [`Task`](https://hexdocs.pm/elixir/Task.html)?

sirn: AFAIK, `Task` is designed as a convenient way to write async code (say, requesting 5 websites simultaneously) rather than for a case like this, where it might require GenServer to spawn more processes or talk with other process (e.g. for the upcoming `Keychain` application).

It’s possible to define task as a module, but that’d require a supervisor and the `Task` module may not be as flexible as a `GenServer` implementation. For example, in the future if we were to switch to having process standby for few minutes until timeout, the current code would pretty much work as-is, where I don’t think it would be the case with Task.

apps/blockchain/lib/blockchain/backend.ex
sirn: It shouldn’t, because we didn’t catch anything. `after` just ensure the code always get executed regardless whether the code in `try` block raises or not.

apps/blockchain/lib/blockchain/backend.ex
unnawut: Makes total sense. Resolving.

PR #812, opened 2019-02-21
Improve configuration validation

### LINKED ISSUE #661: Improve validation for configurations
some fields like `max_per_page` `forget_password_request_lifetime` `max_password` could be null even with integer type validation.

ripzery:
Issue/Task Number: #661

Closes #661

# Overview
This PR prevents the following setting fields to be updated as `nil` or negative integer:

- `max_per_page`
- `min_password_length`
- `forget_password_request_lifetime`
- `balance_caching_reset_frequency`

# Changes

- Change the following settings from `integer` to `unsigned-integer`
    - `max_per_page`
    - `min_password_length`
    - `forget_password_request_lifetime`
    - `balance_caching_reset_frequency`

### DIFF
--- a/apps/admin_api/test/admin_api/v1/controllers/configuration_controller_test.exs
+++ b/apps/admin_api/test/admin_api/v1/controllers/configuration_controller_test.exs
@@ -116,6 +116,98 @@ defmodule AdminAPI.V1.ConfigurationControllerTest do
              }
     end
 
+    test_with_auths "updates nil for all unsigned_integer settings", context do
+      response =
+        request("/configuration.update", %{
+          max_per_page: nil,
+          min_password_length: nil,
+          forget_password_request_lifetime: nil,
+          balance_caching_reset_frequency: nil,
+          config_pid: context[:config_pid]
+        })
+
+      assert response["success"] == true
+      data = response["data"]["data"]
+
+      assert data["max_per_page"] == %{
+               "code" => "client:invalid_parameter",
+               "description" =>
+                 "Invalid parameter provided. `value` must be of type 'unsigned_integer'.",
+               "messages" => %{"value" => ["invalid_type_for_value"]},
+               "object" => "error"
+             }
+
+      assert data["min_password_length"] == %{
+               "code" => "client:invalid_parameter",
+               "description" =>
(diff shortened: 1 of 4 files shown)


unnawut:
Doesn't allowing null means that admins can chose to go back to the system's default values?

unnawut:
@ripzery @T-Dnzt @sirn @jarindr What do you think? To me allowing null allows a user interface that the user could pick a `Use default` option.

T-Dnzt:
@unnawut none of the changed settings have a default `nil` value tho, they all have pre-defined integer values so I think it's fine to have them non-nullable.

apps/admin_api/test/admin_api/v1/controllers/configuration_controller_test.exs
```diff
@@ -109,12 +109,67 @@ defmodule AdminAPI.V1.ConfigurationControllerTest do
...
+        "object" => "error"
+      }
+
+      assert data["max_per_page"] == error
+      assert data["min_password_length"] == error
+      assert data["forget_password_request_lifetime"] == error
+      assert data["balance_caching_reset_frequency"] == error
+    end
+
+    test_with_auths "updates negative integer for all unsigned_integer settings", context do
```
unnawut:
```suggestion
    test_with_auths "returns errors when updating unsigned_integer settings with negative integers", context do
```

apps/admin_api/test/admin_api/v1/controllers/configuration_controller_test.exs
```diff
@@ -109,12 +109,67 @@ defmodule AdminAPI.V1.ConfigurationControllerTest do
...
                "code" => "client:invalid_parameter",
-               "description" => "Invalid parameter provided. `value` must be of type 'integer'.",
+               "description" =>
+                 "Invalid parameter provided. `value` must be of type 'unsigned_integer'.",
                "messages" => %{"value" => ["invalid_type_for_value"]},
                "object" => "error"
              }
     end
 
+    test_with_auths "updates nil for all unsigned_integer settings", context do
```
unnawut:
```suggestion
    test_with_auths "returns errors when updating unsigned_integer settings with nil", context do
```

PR #833, opened 2019-03-08
Fix production image compilation 

sirn:
* Remove Circle CI cache as it’s causing NIF to fail to load.
* Simplify `.circleci/config.yml` by using short form for `run:`
* Simplify `.circleci/config.yml` by splitting long scripts into an actual script file.
* Use `make docker` in CI so we have consistent way to build Docker image.
* Improve conditional CI steps using ENV var detection instead of `CIRCLE_PROJECT_USERNAME`.

### DIFF
--- a/.circleci/ci_deploy.sh
+++ b/.circleci/ci_deploy.sh
@@ -1,0 +1,110 @@
+#!/bin/sh
+
+set -e
+
+echo_info() {
+    printf "\\033[0;34m%s\\033[0;0m\\n" "$1"
+}
+
+echo_warn() {
+    printf "\\033[0;33m%s\\033[0;0m\\n" "$1"
+}
+
+
+## Sanity check
+##
+
+if [ -z "$CIRCLE_GPG_KEY" ] ||
+       [ -z "$CIRCLE_GPG_OWNERTRUST" ] ||
+       [ -z "$GCP_KEY_FILE" ] ||
+       [ -z "$GPG_ACCOUNT_ID" ] ||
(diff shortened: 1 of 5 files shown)


sirn:
Note that removing cache cause the build to take longer, but it should result in a more predictable builds (due to no more randomly failed NIFs loading). Some timing statistics:

**With cache:**

* without E2E: ~6m
* with E2E: ~18m

**Without cache:**

* without E2E: ~8m
* with E2E: ~20m

.circleci/ci_deploy.sh
```diff
@@ -0,0 +1,110 @@
...
+}
+
+
+## Sanity check
+##
+
+if [ -z "$CIRCLE_GPG_KEY" ] ||
+       [ -z "$CIRCLE_GPG_OWNERTRUST" ] ||
+       [ -z "$GCP_KEY_FILE" ] ||
+       [ -z "$GPG_ACCOUNT_ID" ] ||
```
unnawut: Did you mean `$GCP_ACCOUNT_ID`?

.circleci/ci_deploy.sh
```diff
@@ -0,0 +1,110 @@
...
+       [ -z "$GCP_REGION" ] ||
+       [ -z "$GCP_ZONE" ] ||
+       [ -z "$GCS_CLUSTER_ID" ]; then
```
unnawut: `$GCP_CLUSTER_ID`?

sirn:
I’ll cherry-pick this to 1.2

PR #834, opened 2019-03-09
Remove 5-association limit from MatchParser #783

### LINKED ISSUE #783: Remove 5-association limit from MatchParser
Since now we're now on Ecto 3 and [[Ecto.Query] Allow joins in queries to be named via :as and allow named binding](https://github.com/elixir-ecto/ecto/blob/master/CHANGELOG.md). We can now get rid of the clunky [`MatchAllQuery`](https://github.com/omisego/ewallet/blob/master/apps/ewallet/lib/ewallet/web/filters/match_all_query.ex) and [`MatchAnyQuery`](https://github.com/omisego/ewallet/blob/master/apps/ewallet/lib/ewallet/web/filters/match_any_query.ex) workarounds.

Thanks to @jarindr for bringing this up.

enyan94:
Closes #783 

Queries are no longer limited by 5-associations. 

The issue description suggested using named bindings but I found that using the existing Dynamic Queries with dynamic positioning to be better. Named bindings required creating compile-time atoms for every association within the schema in order to work. This added unnecessary complexity.

Implementation outline:
- do_filter_assoc() was generalized in MatchAllQuery and MatchAnyQuery
- checks for association limit was removed from MatchParser; 'position' variable adjusted.
- Added test case march_parser_tests.ex to check for 6-associations; also removed redundant test for  5-association limit


unnawut:
Wow! How did I not know a `{schema, position}` tuple works! For my own knowledge, could you point me to the docs that define this?

unnawut:
> Wow! How did I not know a `{schema, position}` tuple works! For my own knowledge, could you point me to the docs that define this?

Apologies for asking what you already mentioned in #783. Too many things on my head today 🙇

unnawut:
I think we can bite the bullet on this one and use their internal API. But one last thing before doing that is to document the rationale for using the internal API and a game plan in case this internal API breaks, right into the code. E.g. that we could revert back to the hardcoded workaround, and also put a reference link back to this PR and its Github issue.

@enyan94 Could you help add this the code comment for both modules?

Then I think this is good to go! 🎉 

---

In parallel, we could also submit a proposal to either promote this `{schema, position}` into a public API or have Ecto support dynamic named bindings, as suggested by Jose in https://github.com/elixir-ecto/ecto/issues/2832. This should be a pretty valid use case, but shouldn't need to block this PR. @enyan94 Let me know as well if you'd like to do this.

unnawut:
Added more reviewers here in case they have differing opinions regarding using Ecto's internal API.

enyan94:
@unnawut I added the comments like you suggested. I'm also open to submitting a proposal. Would that basically be creating an issue on the Ecto repo requesting for this feature?

unnawut:
> Would that basically be creating an issue on the Ecto repo requesting for this feature?

@enyan94 Actually you'll need to email https://groups.google.com/forum/#!forum/elixir-ecto and prefix the subject with `[Proposal]`. I tried to search just now but hasn't seen any similar proposal yet. So please go ahead!

enyan94:
Okay. The proposal has been submitted. [Link](https://groups.google.com/forum/#!topic/elixir-ecto/cs_8XinSvu0)

unnawut:
@enyan94 Thank you! One very last thing. Our `v1.2` branch is due for earlier release than `master` so I suggest we base this PR off `v1.2`.

I couldn't do that without destroying your authorship on the commits so could you help change the base to `v1.2`?

enyan94:
@unnawut Oh, sorry.  I didn't know we were using branch v1.2.

It's including other commits though? Is that okay?

unnawut:
Unfortunately you'll have to rebase without those extra commits. Master is not near a release yet. A `git rebase --interactive` is probably the easiest way. Sorry the trouble!

enyan94:
@unnawut Sorry, I couldn't properly rebase it :/ 

I had to create a new branch and [PR](https://github.com/omisego/ewallet/pull/858)

I hope that's okay.

unnawut:
Sounds good! Thank you!

Closing this one then.

PR #850, opened 2019-03-13
fix: env variables

InoMurko:

### DIFF
--- a/apps/status/lib/status.ex
+++ b/apps/status/lib/status.ex
@@ -70,12 +70,9 @@ defmodule Status do
 
   @spec is_enabled?() :: boolean() | nil
   defp is_enabled?() do
-    case {Application.get_env(:status, :metrics), Normalize.to_boolean(System.get_env("METRICS"))} do
-      {_, true} -> true
-      {_, false} -> false
-      {true, _} -> true
-      {false, _} -> false
-      _ -> nil
-    end
+    Normalize.to_boolean(System.get_env("METRICS"))
+  rescue
+    Normalize.ToBooleanError ->
+      Application.get_env(:status, :metrics)
   end
 end
--- a/apps/status/lib/status.ex
+++ b/apps/status/lib/status.ex
@@ -21,6 +21,7 @@ defmodule Status do
   alias Utils.Helpers.Normalize
 
   def start(_type, _args) do
+    DeferredConfig.populate(:status)
     import Supervisor.Spec, warn: false
 
     metrics =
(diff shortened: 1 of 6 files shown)


apps/status/lib/status.ex
```diff
@@ -70,12 +70,9 @@ defmodule Status do
...
-      {_, true} -> true
-      {_, false} -> false
-      {true, _} -> true
-      {false, _} -> false
-      _ -> nil
-    end
+    Normalize.to_boolean(System.get_env("METRICS"))
+  rescue
+    Normalize.ToBooleanError ->
+      Application.get_env(:status, :metrics)
```
T-Dnzt: Hmm, I don't really see the need for this. How about something like:

```
  defp is_enabled?() do
    :status
    |> Application.get_env(:metrics)
    |> Normalize.to_boolean(()
  end
```

And having the `metrics` value set in the config files:

prod
```
config :status,
  metrics: {:system, "METRICS"}
```

I'm not sure if this up should have `false` as default as well, the whole thing stays optional, right? 

dev
```
config :status,
  metrics: {:system, "METRICS", false}
```

test
```
config :status,
  metrics: {:system, "METRICS", false}
```

I think this is the pattern we follow for most of our ENVs.

T-Dnzt: Could also just put it in the main config file if we use `false` as default everywhere.

apps/utils/lib/helpers/path_resolver.ex
```diff
@@ -25,7 +25,7 @@ defmodule Utils.Helpers.PathResolver do
   path from a file.
   """
   def static_dir(app) do
-    serve_local_static = System.get_env("SERVE_LOCAL_STATIC")
+    serve_local_static = System.get_env("SERVE_LOCAL_STATIC") || false
```
T-Dnzt: I'm wondering, should the default value actually be `true` or `false`?

apps/status/lib/status.ex
InoMurko: I've noticed that (at least Phoenix) is moving away from this approach: https://github.com/phoenixframework/phoenix/blob/70606bf9e98b0a838ccea1d5e376d4dec78c446b/lib/phoenix/socket/transport.ex#L506

Does this apply here?

T-Dnzt: Hmm, I think we went that way because of Deferred Config (I could be wrong) 🤔 If we decide to move away from this, it should be done globally throughout the project. For now, let's keep it consistent. Thanks for raising it tho, didn't know Phoenix was moving away from it.

apps/utils/lib/helpers/path_resolver.ex
InoMurko: No idea.

InoMurko: It was false before, though. Normalize.to_boolean/1 changed, thats why we required to be explicit.

T-Dnzt: Okay, let's leave it at `false` then! Thanks

PR #863, opened 2019-03-17
Persist generated private key in Keychain DB

sirn:
Issue/Task Number: #689

### DIFF
--- a/apps/blockchain_eth/lib/blockchain_eth/worker.ex
+++ b/apps/blockchain_eth/lib/blockchain_eth/worker.ex
@@ -25,6 +25,7 @@ defmodule BlockchainEth.Worker do
   ##
 
   use GenServer
+  alias Keychain.Key
   alias ExthCrypto.ECIES.ECDH
   alias ExthCrypto.Hash.Keccak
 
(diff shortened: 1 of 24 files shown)


apps/keychain/config/prod.exs
```diff
@@ -0,0 +1,6 @@
+use Mix.Config
+
+config :keychain, Keychain.Repo,
+  adapter: Ecto.Adapters.Postgres,
+  url: {:system, "DATABASE_URL", "postgres://localhost/ewallet_prod"},
```
unnawut:
```suggestion
  url: {:system, "DATABASE_URL", "postgres://localhost/ewallet"},
```

apps/keychain/test/keychain/key_test.exs
```diff
@@ -0,0 +1,55 @@
...
+    test "inserts a new private key" do
+      assert Repo.all(Key) == []
+      {:ok, key} = Key.insert_private_key("key-1", "private-key-1")
+      assert Repo.all(Key) == [key]
+    end
+
+    test "raises if wallet id already exists" do
+      key_1 = insert(:key)
+
+      assert_raise ConstraintError, fn ->
```
unnawut: Should this return `{:error, changeset}` instead?

apps/blockchain_eth/lib/blockchain_eth/worker.ex
```diff
@@ -73,7 +74,10 @@ defmodule BlockchainEth.Worker do
       |> Keccak.kec()
       |> Base.encode16(case: :lower)
 
-    {:reply, {:ok, "0x#{account_encoded}", pkey_encoded}, reg}
+    wallet_id = "0x#{account_encoded}"
+
+    {:ok, _} = Key.insert_private_key(wallet_id, pkey_encoded)
```
T-Dnzt: This should be storing the private key, not the public one 😅

apps/keychain/test/keychain/key_test.exs
sirn: I don't think so, this is a "should never happen" scenario and I think it's better to be hard fail than a soft fail.

PR #871, opened 2019-03-22
Fix cors config

### LINKED ISSUE #866: Fix CORS bug

mederic-p:
Issue/Task Number: 866
Closes #866 

# Overview

This PR is an attempt to fix the currently broken CORS plug.
The previous configuration was not working properly because the configuration was set in application variables after the plug was initialized.

This PR moves the plug configuration setup at the `endpoint` level.

# Changes

- Configure the plug at the `endpoint` level
- Refactor plug config

# Impact

As the plug was not properly working until now, it was accepting all origins by default.
After this PR gets merged a proper configuration will be required.


unnawut:
Since you're already on this, can you try turning the ENV var reference into a deferred config? 🙏

PR #875, opened 2019-03-22
Normalize ENV values before saving to settings

### LINKED ISSUE #804: Settings migration should normalize input before saving
* `redirect_url_prefixes` needs to be an array
* `enable_standalone` needs to be boolean

unnawut:
Issue/Task Number: #804
Closes #804

# Overview

This PR normalizes an ENV value to its respective settings' type before attempting to update into settings.

# Changes

- Add `Config.update_one/3` for convenience when wanting to update a single settings
- Add `Normalize.to_strings/1` to normalize a comma-separated string (as expected as ENV values) into a list of strings.
- Updated `EWallet.ReleaseTasks.Config` with ENV value normalization according to each setting's type
- Updated `EWallet.ReleaseTasks.ConfigMigration` with ENV value normalization according to each setting's type

# Implementation Details

Notice that there's duplicate normalization code in `ReleaseTasks.Config` and `ReleaseTasks.ConfigMigration`. I think this is very specific to normalizing ENV values, which is unlikely to be used anywhere else, so it's simpler to duplicate code here than having another shared module for this.

# Usage

```shell
$ REDIRECT_URL_PREFIXES=a,b,c ENABLE_STANDALONE=1 mix omg.config -m -y
...
  - Skipped: `redirect_url_prefixes` is already ["a", "b", "c", "d"].
  - Migrated: `enable_standalone` is now true.
...
```

```shell
$ mix omg.config redirect_url_prefixes a,b,c,d
...
Updated: `redirect_url_prefixes` is now ["a", "b", "c", "d"].
```

```shell
$ mix omg.config enable_standalone true
...
Updated: `enable_standalone` is now false.
```

```shell
$ mix omg.config enable_standalone true
...
Skipped: `enable_standalone` is already set to true.
```

# Impact

Some setups may not have all ENV values migrated to Settings yet due to this issue. We recommend running `bin/ewallet config --migrate` once more.

No changes to API specs nor DB schemas.

### DIFF
--- a/apps/ewallet/lib/ewallet/release_tasks/config.ex
+++ b/apps/ewallet/lib/ewallet/release_tasks/config.ex
@@ -99,16 +99,17 @@ defmodule EWallet.ReleaseTasks.Config do
 
       %{value: existing, type: type} ->
         case cast_env(value, type) do
-          ^existing ->
-            {:unchanged, existing}
-
-          casted_value ->
-            # The GenServer will return {:ok, result}, where the result is the actual
-            # {:ok, _} | {:error, _} of the operation, so we need to return only the result.
-            {:ok, result} = Config.update(%{key => casted_value, :originator => %CLIUser{}})
-            Enum.find_value(result, fn {^key, v} -> v end)
+          ^existing -> {:unchanged, existing}
+          casted_value -> do_update(key, value)
         end
     end
+  end
+
+  defp do_update(key, value) do
+    # The GenServer will return {:ok, result}, where the result is the actual
+    # {:ok, _} | {:error, _} of the operation, so we need to return only the result.
+    {:ok, result} = Config.update(%{key => value, :originator => %CLIUser{}})
+    Enum.find_value(result, fn {^key, v} -> v end)
   end
 
   # These cast_env/2 are private to this module because the only other place that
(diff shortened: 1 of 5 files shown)


apps/utils/lib/helpers/normalize.ex
```diff
@@ -72,4 +72,11 @@ defmodule Utils.Helpers.Normalize do
   def to_integer(s) when is_integer(s), do: s
   def to_integer(s) when is_float(s), do: :erlang.round(s)
   def to_integer(value), do: raise(ToIntegerError, message: ToIntegerError.error_message(value))
+
+  # Converts a comma-separated string to a list of strings
+  def to_strings(s) do
```
T-Dnzt: 👍

unnawut:
Good point, I'll update the logic to prepare all the changes first then update_all + reload just once.

apps/ewallet/lib/ewallet/release_tasks/config.ex
```diff
@@ -60,17 +61,62 @@ defmodule EWallet.ReleaseTasks.Config do
...
+
+  defp do_set_config(key, value) do
+    case Config.get_setting(key) do
+      nil ->
+        {:error, :setting_not_found}
+
+      %{value: existing, type: type} ->
+        case cast_env(value, type) do
+          ^existing -> {:unchanged, existing}
+          casted_value -> do_update(key, value)
```
T-Dnzt:
```
warning: variable "casted_value" is unused (if the variable is not meant to be used, prefix it with an underscore)
  lib/ewallet/release_tasks/config.ex:103
```

T-Dnzt: Shouldn't you use the `casted_value` btw?

unnawut: 🤦‍♂️

PR #883, opened 2019-03-25
Admin panel assigning key to account

jarindr:
Issue/Task Number: #882 

# Overview

With the new feature we have in 1.2,  we need to be able to assign key to account and vice versa.

# Changes

- Update design for creating keys.
- Renaming api_key and access_key to admin_key and client_key
- Refactoring around.
- New services for assigning keys.
- Handle recent account load abnormally when switching branches and reseeding.

# Usage

- Create a key in manage section, both admin and client.
- try to assign a key to account section.

# Impact

Deploy as usual.


apps/admin_panel/assets/src/omg-page-api/AdminKeySection.js
```diff
@@ -0,0 +1,288 @@
...
+        open={this.state.privateKeyModalOpen}
+        onRequestClose={this.onRequestCloseShowPrivateKey}
+        onOk={this.onRequestCloseShowPrivateKey}
+        confirmText='Got it!'
+        cancel={false}
+      >
+        <ConfirmCreateKeyContainer>
+          <h4>Your key pair</h4>
+          <p>
+            Please copy and keep this pair of acesss and secret key. Secret key will use to open
```
nicholasmueller: "Please keep a copy of these secret keys. The secret key will be used to access your encrypted information." -> maybe better wording

apps/admin_panel/assets/src/omg-app-layout/ProfileDropdown.js
```diff
@@ -71,7 +71,7 @@ const enhance = compose(
     { logout }
   )
 )
-class ProfileAvatarDropdown extends Component {
+class ProfileAvstyled(atarDropdown)s Component {
```
nicholasmueller: typo

apps/admin_panel/assets/src/omg-access-key/action.test.js
```diff
@@ -56,8 +56,8 @@ describe('apikeys actions', () => {
         data: 'key'
       }
     ]
-    return store.dispatch(createAccessKey()).then(() => {
-      expect(accessKeyService.createAccessKey).toBeCalled()
+    return store.dispatch(createAccessKey({ name: 'PRAYUTHHHHHHHHHHHHHHH' })).then(() => {
```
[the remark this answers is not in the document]
jarindr: #savethailand

apps/admin_panel/assets/src/omg-app-layout/ProfileDropdown.js
jarindr: 🆗

T-Dnzt:
Global level key page
- [ ] Role sorting (Super Admin -> Admin -> Viewer -> None)
- [ ] Top notification text: New admin key successfully created.
- [ ] Better text for key creation confirmation (see at the bottom)
- [ ] Table column order (label before key?)
- [ ] ADMIN KEY -> ACCESS KEY (in colum names)
- [ ] Status column should at the end
- [ ] created date -> Creation Date or Created At (I prefer Created At, let's keep it consistent with the rest)
- [ ] Humanize global role names: Super Admin, not super_admin
- [ ] Going with full page modals for everything now? Feels like small forms could keep small modals
- [ ] No search field?
- [ ] Add download csv? How hard is it to generate a csv file in JavaScript so that users can download the details of their keys (like AWS)

Account level key page
- [ ] Assign Key -> should be Assign or Create Key OR we should have another button to create a key
- [ ] Add creation date in the list of keys when trying to assign one if possible (in the dropdown list)
- [ ] filtering doesn't work for name (not sure if backend or frontend)
- [ ] Admin above viewer
- [ ] Why role is capitalized in dropdown but not in field?
- [ ] Remove "optional" for account, it's not
- [ ] Can we filter keys that are already assigned from the dropdown? Frontend filtering is fine
- [ ] Capitalize the role in the table, put name at the beginning, change created date, move status at the end
- [ ] Search not working with access key value, only name

Backend
- [ ] make name mandatory for keys and api keys from now on?
- [ ] Add creator_key or creator_user to key serializer so we can show those in dropdown list

---
Your key pair
Please keep a copy of these secret keys. The secret key will be used to access your encrypted information.
---
change to
---
An access key and a secret have been generated. In order to access the admin API programmatically, you will need to use them together.
We do not store the secret key in our database for security reasons. This is the only time you will see it, so copy it and store it somewhere safe.
Note that you will be able to get the access key anytime from the admin panel.  
---

PR #896, opened 2019-03-27
Validate master_account setting

### LINKED ISSUE #840: Validate the master_account setting

T-Dnzt:
Issue/Task Number: #840 
closes #840

# Overview

This PR ensures that only a valid account ID can be set through the Admin API.

# Changes

- Do pre-validation in the configuration controller. This validation can't go in the config sub app because it involves checking than an account exists (the config subapp doesn't know/care about the eWallet DB entities).


apps/admin_api/lib/admin_api/v1/controllers/configuration_controller.ex
```diff
@@ -37,13 +39,51 @@ defmodule AdminAPI.V1.ConfigurationController do
...
 
+  defp validate_master_account(%{"master_account" => master_account_id} = attrs) do
+    case Account.get(master_account_id) do
+      nil ->
+        {
+          Map.delete(attrs, "master_account"),
+          {
+            :master_account,
+            {:error,
+             %Changeset{
```
unnawut:
```elixir
%StoredSetting{}
|> Changeset.change(%{value: master_account_id})
|> Changeset.apply_action(:update)
|> Changeset.add_error(:value, "must match an existing account", validation: :account_not_found)
```

... so we don't need to handle future Ecto's implementation changes.

Also updated `validation: master_account_id` to `validation: :account_not_found` so it aligns with other `add_error/4` usage (`validation: :validation_atom`).

T-Dnzt: But using `validation: :account_not_found` returns 

```
 %{"code" => "client:invalid_parameter", "description" => "Invalid parameter provided. `value` must match an existing account.", "object" => "error", "messages" => %{"value" => ["account_not_found"]}}
```

While I wanted the `"value"` to have the actual value that was sent by the user 🤔

unnawut: The other validators already interpolated value into the error description string though: https://github.com/omisego/ewallet/blob/v1.2/apps/ewallet_config/lib/ewallet_config/setting_validator.ex

Issue #890, opened 2019-03-26
Admin Panel UI Improvements "Easy"


nicholasmueller:
* Search bar component could look cleaner/have a cleaner transition. ie. don’t transition X icon from the left, keep X in place. Would be nice to debounce the search, requiring onEnter is not that obvious to the user
* I think dropdown arrow and breadcrumb arrows are a little bit too big
* Accounts -> Create account modal -> add to category -> checkbox not aligned with category name
* Accounts -> “Account successfully created” avatar defaults to empty grey box, should we use the 3 character generated avatar
* Tokens -> Don’t allow negative input on the “mint token” modal
* Configuration -> After making a change and not saving, user can navigate to another page… user should be prompted with “You have unsaved changes, etc” -> use React Router Prompt component for quick solution
* Requests Overview -> Slide transition the side modal
* My Profile -> “Logout” hover cursor
* Activity Logs -> Search bar placeholder text is cutoff
* Be able to close all modals onblur

Issue #888, opened 2019-03-26
Admin Panel Bugs


nicholasmueller:
* Accounts -> If I manually go to http://localhost:9000/admin/accounts/RANDOMACCOUNT/detail it will create a non existent account on sidebar,
* Accounts -> Clicking “User Wallets” dropdown, then clicking another tab, then clicking “User Wallets” again, will incorrectly show “Account Wallets” screen or not update the tab title
* Activity Logs -> Clicking “originator” in detail sometimes points to wrong url address

PR #901, opened 2019-03-28
UI improvements

nicholasmueller:
Issue/Task Number: [890](https://github.com/omisego/ewallet/issues/890)
also closes [888](https://github.com/omisego/ewallet/issues/888)

# Overview

Made minor changes cleaning up the ui 
also fixed all bugs in issue 888

# Changes

- Cleaned up the searchbar opening animation. Also added a 500ms debounce to the search so the user doesn't have to hit 'enter' but still can.
- some CSS tweaks
- Block the negative keypress on mint token modal
- Prompt if unsaved changes on the configuration page when navigating away
- Slide transition for the side panels.
- Accordion transition for advancedsettings in createtransactionrequest
- New exchangepair modal design

# Usage

- go to overview tables and click a row to see the new side transition
- play with the searchbar
- make a change on the configuration page and try navigate to another page without clicking save
- play with "Create Exchange Pair" modal on Tokens page
- open advancedsettings in createtransactionrequest


apps/admin_panel/assets/src/omg-exchange-rate-modal/index.js
```diff
@@ -169,17 +209,37 @@ class CreateExchangeRateModal extends Component {
...
+    } = this.state
+
+    const forwardRate = _.round(toTokenRate / fromTokenRate, 3)
+    const backRate = _.round(1 / forwardRate, 3)
 
     return (
       <>
-        <div>
-          {`1 ${this.state.fromTokenSearch} : ${forwardRate} ${this.state.toTokenSearch}`}
+        <div id='calculation-title'>Exchange Pair</div>
```
jarindr: let's keep `id` for the super hacky stuffs, classname as default first

apps/admin_panel/assets/src/omg-exchange-rate-modal/index.js
```diff
@@ -169,17 +209,37 @@ class CreateExchangeRateModal extends Component {
...
+        <RateContainer>
+          <div>
+            {`1 ${fromTokenSymbol} / ${forwardRate} ${toTokenSymbol}`}
+          </div>
+          <BackRateContainer disabled={onlyOneWayExchange}>
+            {`1 ${toTokenSymbol} / ${backRate} ${fromTokenSymbol}`}
+          </BackRateContainer>
+        </RateContainer>
+
+        <div id='calculation-disclaimer'>
```
jarindr: here too

apps/admin_panel/assets/src/omg-exchange-rate-modal/index.js
jarindr: let's fix another bug too 😢 
- when submitting, need to extract out the real number before sending it. For now it works only if amount is < 999 ( no ,)

PR #907, opened 2019-03-29
Wallet balances permission check

T-Dnzt:
Issue/Task Number: #861 
#861

# Overview

This PR removes the balance loading from the `WalletSerializer` and moves it to the `WalletController` with a permission check. 

# Changes

- Move balance loading to controller
- Check permissions for `view_balance` for scope queries and single record loading
- Update bouncer to handle `view_balance` scope permission check

# Implementation Details

Balances will only be set in the serializer if they've already been set previously on the wallet.

# Impact

We might end up with some endpoints not behaving like before (balances not being included anymore).


apps/admin_api/lib/admin_api/v1/controllers/wallet_controller.ex
```diff
@@ -112,7 +111,7 @@ defmodule AdminAPI.V1.WalletController do
          {:ok, _} <- authorize(:enable_or_disable, conn.assigns, wallet),
          attrs <- Originator.set_in_attrs(attrs, conn.assigns),
          {:ok, updated} <- Wallet.enable_or_disable(wallet, attrs),
-         {:ok, updated} <- BalanceLoader.add_balances(updated) do
+         updated <- add_balances_if_allowed(updated, conn) do
```
unnawut: Won't this without `{:ok, updated}` let `{:error, code, description}` below slip into the happy path?

apps/admin_api/lib/admin_api/v1/controllers/wallet_controller.ex
```diff
@@ -123,6 +122,26 @@ defmodule AdminAPI.V1.WalletController do
...
+         true <- !is_nil(query) do
+      BalanceLoader.add_balances_to_wallets_intersect(paginator, query)
+    else
+      _ -> paginator
+    end
+  end
+
+  defp add_balances_if_allowed(wallet, conn) do
+    with {:ok, _} <- authorize(:view_balance, conn.assigns, wallet) do
+      {:ok, wallet} = BalanceLoader.add_balances(wallet)
```
unnawut: Should this go inside the `with` clause? Noticed `BalanceFetcher` will always return `{:ok, _}` but future-proofing...

apps/ewallet/lib/ewallet/policies/policy_helper.ex
```diff
@@ -19,6 +19,10 @@ defmodule EWallet.PolicyHelper do
   alias EWallet.{Bouncer, Bouncer.Permission}
   alias Ecto.Query
 
+  def authorize(:view_balance, attrs, type, schema, nil) do
+    authorize_scope(:export, attrs, type, schema)
```
unnawut:
```suggestion
    authorize_scope(:view_balance, attrs, type, schema)
```

PR #920, opened 2019-04-03
Implement 2FA Authentication

### LINKED ISSUE #533: Add 2FA support via authenticator

ripzery:
Issue/Task Number: #533 

Closes #533 

# Overview

This PR adds support for two factors authentication. The following list are new endpoints:

- `/admin.login_2fa`
- `/me.create_backup_codes`
- `/me.create_secret_code`
- `/me.enable_2fa`
- `/me.disable_2fa`

# Changes

**New:**

1. **TwoFactorAuthenticationController**
  - Handles request and response

2. **TwoFactorAuthenticator**
  - Handles `enable`, `disable`
  - Delegates `create` and `verify` call to appropriated authenticator which can be one of `BackupCodeAuthenticator` or `PasscodeAuthenticator`

3. **BackupCodeAuthenticator**
  - Handles `create, verify` for backup code.

4. **PasscodeAuthenticator**
  - Handles `create, verify` for app generated code.

5. **PreAuthToken**
 - A temporary token, which will be used for calling `/admin.login_2fa`.

6. **UserAuthToken**
- Handle user authentication to get whether `AuthToken` or `PreAuthToken`.

# Implementation Details

## Secret and Backup Codes Generation

- Secret

```elixir
iex> 10 |> :crypto.strong_rand_bytes() |> Base.encode32()
"HA4ZUTBQDHALT7GR"
```

- Backup Codes

```elixir
iex> 4 |> :crypto.strong_rand_bytes() |> Base.encode16(case: :lower)
"8a011a72"

iex> 1..10 |> Enum.map(fn _ -> 4 |> :crypto.strong_rand_bytes() |> Base.encode16(case: :lower) end)
["57e31848", "11ac987e", "9d4d88fd", "36594326", "dce9e4b1", "bec5b9ff",
d2643b97", "0e300670", "8aa0d753", "b19b9304"]
```

## Migration

**User**

- field(:enabled_2fa_at, :naive_datetime)
- field(:secret_2fa_code, :string)
- field(:encrypted_backup_codes, {:array, :string}, default: [])

New table **PreAuthToken**.
# Usage

## To enable 2FA
  - Call `/admin.login` to log in with username and password.
  - Call `/me.create_secret_code` with empty boydy. (Use the response to create the 2FA QR code)
  - Call `/me.create_backup_codes` with empty body.
  - Call `/me.enable_2fa` with the following body:
```json
{
  "passcode": "123456",
  "backup_code": "12345678"
}
```

> Now all authentication tokens belong to the user was deleted.

## To disable 2FA
  - Call `/admin.login` to log in with username and password.
  - Call `/admin.login_2fa` with the following body (need auth token from previous `/admin.login`):
```json
{
  "passcode": "123456"
}
```
or
```json
{
  "backup_code": "12345678"
}
```
 - Call `/me.disable_2fa` with the same body as `/admin.login_2fa`

> Now all authentication tokens belong to the user was deleted.

## To log in with 2FA
  - Call `/admin.login` to log in with username and password.
  - Call `/admin.login_2fa` with the following body (need auth token from previous `/admin.login`):
```json
{
  "passcode": "123456"
}
```
or
```json
{
  "backup_code": "12345678"
}
```
> Now all authentication tokens and pre-authentication tokens belong to the user was deleted, except one, the auth token that returned from the `/admin.login_2fa`.

# Impact

Need `mix do ecto.create, ecto.migrate`.


unnawut:
Can you add user flows, i.e. what endpoint to call one after another as Step 1, 2, 3 for these scenarios?
  - To enable 2FA
  - To login with 2FA
  - To disable 2FA

unnawut:
1. Regarding **To login with 2FA** in the description. Does it mean that `/admin.login` will authenticate the user even without 2FA? Because the authentication token is already always returned during `/admin.login`?

2. Regarding **To disable 2FA** in the description, is calling `/2fa.create_secret_code` really needed?

apps/ewallet/lib/ewallet/auth/backup_code_authenticator.ex
```diff
@@ -0,0 +1,83 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+alias Utils.Helpers.Crypto
+
+defmodule EWallet.BackupCodeAuthenticator do
+  @moduledoc """
```
unnawut: Great docs!

apps/ewallet/lib/ewallet/auth/backup_code_authenticator.ex
```diff
@@ -0,0 +1,83 @@
...
+  {:ok}
+
+  # Failure
+  {:error, :invalid_backup_code}
+
+  """
+  @number_of_bytes 4
+
+  def verify(encrypted_backup_codes, backup_code) do
+    if Enum.any?(encrypted_backup_codes, &Crypto.verify_password(backup_code, &1)) do
```
unnawut: 👏👏👏👏👏👏👏👏👏 for using `Crypto.verify_password/2`

apps/ewallet/lib/ewallet/auth/backup_code_authenticator.ex
```diff
@@ -0,0 +1,83 @@
...
+      {:ok}
+    else
+      {:error, :invalid_backup_code}
+    end
+  end
+
+  def create(number_of_backup_codes) do
```
unnawut: Great modularity

apps/ewallet/lib/ewallet/auth/backup_code_authenticator.ex
```diff
@@ -0,0 +1,83 @@
...
+    backup_codes =
+      1..number_of_backup_codes
+      |> Enum.map(fn _ -> do_create() end)
+
+    encrypted_backup_codes =
```
unnawut:
```suggestion
    hashed_backup_codes =
```

apps/ewallet/lib/ewallet/auth/two_factor_authenticator.ex
```diff
@@ -0,0 +1,141 @@
...
+  - enable:
+    - Update `required_2fa` to `true` for all auth tokens belong to this user. (all tokens cannot be used anymore.)
+    - Update `enabled_2fa_at` to the current naive date time for user entity.
+    - Generate new auth_token with `required_2fa` `false`. (this token is 2FA authenticated. Use this token instead.)
+  - disable:
+    - Clear `enabled_2fa_at`, `encrypted_backup_codes`, `secret_2fa_code` for user entity.
+    - Update `required_2fa` to `false` for the current auth token. (this token can still be used)
+  """
+
+  @number_of_backup_codes 10
```
unnawut: Can also make this configurable through Config

apps/ewallet/lib/ewallet/auth/two_factor_authenticator.ex
```diff
@@ -0,0 +1,141 @@
...
+          "originator" => user
+        }
+      )
+
+    AuthToken.generate(updated_user, owner_app, updated_user, true)
+  end
+
+  # Disable 2FA
+  def enable(%User{} = user, token_string, owner_app, false) do
+    with {:ok, updated_user} <-
```
unnawut: Can potentially cause a limbo. Need to wrap these 3 updates in a transaction.

apps/ewallet/lib/ewallet/auth/two_factor_authenticator.ex
```diff
@@ -0,0 +1,141 @@
...
+          "enabled_2fa_at" => NaiveDateTime.utc_now(),
+          "originator" => user
+        }
+      )
+
+    AuthToken.generate(updated_user, owner_app, updated_user, true)
+  end
+
+  # Disable 2FA
+  def enable(%User{} = user, token_string, owner_app, false) do
```
unnawut:
```suggestion
  def disable(%User{} = user, token_string, owner_app, false) do
```

apps/ewallet/lib/ewallet/auth/two_factor_authenticator.ex
```diff
@@ -0,0 +1,141 @@
...
+           User.set_encrypted_backup_codes(user, %{
+             "encrypted_backup_codes" => [],
+             "originator" => updated_user
+           }),
+         {:ok, updated_user} <-
+           User.set_enabled_2fa_at(user, %{
+             "enabled_2fa_at" => nil,
+             "originator" => updated_user
+           }),
+         :ok = AuthToken.set_required_2fa_for_user(updated_user, false) do
```
unnawut:
```suggestion
         :ok <- AuthToken.set_required_2fa_for_user(updated_user, false) do
```

apps/ewallet_db/lib/ewallet_db/auth_token.ex
```diff
@@ -100,19 +101,30 @@ defmodule EWalletDB.AuthToken do
...
       originator: originator
     }
 
     insert(attrs)
   end
 
-  def generate(_, _, _), do: {:error, :invalid_parameter}
+  def generate(_, _, _, _), do: {:error, :invalid_parameter}
+
+  defp is_required_2fa(%User{enabled_2fa_at: nil}, _is_2fa_token) do
```
unnawut:
```suggestion
  defp required_2fa?(%User{enabled_2fa_at: nil}, _is_2fa_token) do
```

apps/ewallet_db/lib/ewallet_db/auth_token.ex
```diff
@@ -218,6 +233,18 @@ defmodule EWalletDB.AuthToken do
     :ok
   end
 
+  def set_required_2fa_for_user(user, required_2fa) do
```
unnawut: I'm a bit scared of manipulating `AuthToken` and potentially having `AuthToken` that may and may not have permissions.

Shall we just delete all auth_tokens of the user when enabling 2FA?

What do you think? @ripzery @T-Dnzt @sirn

apps/ewallet_db/lib/ewallet_db/auth_token.ex
```diff
@@ -100,19 +101,30 @@ defmodule EWalletDB.AuthToken do
...
   """
-  def generate(%User{} = user, owner_app, originator) when is_atom(owner_app) do
+  def generate(user, owner_app, originator, is_2fa_token \\ false)
+
+  def generate(%User{} = user, owner_app, originator, is_2fa_token) when is_atom(owner_app) do
     attrs = %{
       owner_app: Atom.to_string(owner_app),
       user_uuid: user.uuid,
       account_uuid: nil,
+      required_2fa: is_required_2fa(user, is_2fa_token),
```
unnawut: From the current flow in the PR description, a 2FA-enabled user will have an auth_token before 2FA is verified. A bit scary to me. 

Can we ask for 2FA token to be passed during `/admin.login` and the server returns an auth_token only when username/password/2FA all match?

---

What do you think? @ripzery @T-Dnzt @sirn

apps/ewallet_db/lib/ewallet_db/auth_token.ex
T-Dnzt: Yeah, I like the idea of logging out the users from any current session by deleting the existing tokens.

ripzery:
> 1. Regarding **To login with 2FA** in the description. Does it mean that `/admin.login` will authenticate the user even without 2FA? Because the authentication token is already always returned during `/admin.login`?
> 2. Regarding **To disable 2FA** in the description, is calling `/2fa.create_secret_code` really needed?

1. The authentication token that returned from the `/admin.login` for 2fa-enabled user can be used for only endpoint `/2fa.verify` otherwise, it will return auth error.

2. Sorry, I wrote the wrong description 🤦‍♂️I have corrected it.

apps/ewallet_db/lib/ewallet_db/auth_token.ex
ripzery: @unnawut 

I like this way 👍 The current flow is 2-step verification, but it feels a bit scary for me too because having to manipulate the authentication token. (similar to go.exchange's flow)

I agree with passing the 2FA token to `/admin.login`. However, I have to choose the error response when the enabled-2fa user login with correct username/password but missing 2FA authentication token, what is a more appropriate error code between:
- `user:invalid_login_credentials`: 
 😃More secure by not giving a hint that which field is wrong.
 😟A bit less dynamic login flow.

- `user:some_2fa_error`:
 😃More dynamic login flow, better UX.
 😟A bit less secure by giving a hint that 2FA is required/wrong.
 
Personally, I go for the first `user:invalid_login_credential` because the user should usually know whether they have enabled 2FA or not and we already don't give any clue about `username` and `password` when they are wrong.

ripzery: I like this way :+1: The current flow is 2-step verification, but it feels a bit scary for me too because having to manipulate the authentication token.

I agree with passing the 2FA token to /admin.login. However, I have to choose the error response when the enabled-2fa user login with correct username/password but missing 2FA authentication token, what is a more appropriate error code between:

user:invalid_login_credentials:
:smiley:More secure by not giving a hint that which field is wrong.
:worried:A bit less dynamic login flow.

user:some_2fa_error:
:smiley:More dynamic login flow, better UX.
:worried:A bit less secure by giving a hint that 2FA is required/wrong.

Personally, I go for the first user:invalid_login_credential because the user should usually know whether they have enabled 2FA or not and we already don't give any clue about username and password when they are wrong.

apps/ewallet/lib/ewallet/auth/two_factor_authenticator.ex
```diff
@@ -0,0 +1,196 @@
...
+
+  @default_number_of_backup_codes 10
+  @default_issuer "OmiseGO"
+
+  def login(_, _, %User{enabled_2fa_at: nil}), do: {:error, :user_2fa_disabled}
+
+  def login(attrs, owner_app, %User{} = user) do
+    with {:ok} <- verify(attrs, user),
+         :ok <- PreAuthToken.delete_for_user(user),
+         :ok <- AuthToken.delete_for_user(user) do
```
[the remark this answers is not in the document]
ripzery: Good point 👍 Will figure out how to fix it

apps/ewallet/lib/ewallet/auth/two_factor_authenticator.ex
```diff
@@ -0,0 +1,196 @@
...
+  #
+  # :ok
+  def verify_multiple(attrs, user) when map_size(attrs) > 0 do
+    errors =
+      attrs
+      |> Enum.map(fn {k, v} -> Map.put(%{}, k, v) end)
+      |> Enum.map(&verify(&1, user))
+      |> Enum.filter(&match?({:error, _}, &1))
+
+    if Enum.empty?(errors) do
```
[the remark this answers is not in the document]
ripzery: how to get errors? 🤔

apps/ewallet/lib/ewallet/auth/two_factor_authenticator.ex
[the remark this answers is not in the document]
ripzery: Found that just don't delete auth tokens would be fine because all existed auth tokens were deleted when enabled anyway.

apps/ewallet/lib/ewallet/auth/backup_code_authenticator.ex
```diff
@@ -0,0 +1,94 @@
...
+  def verify(_, nil), do: {:error, :invalid_backup_code}
+
+  def verify(hashed_backup_codes, backup_code)
+      when is_list(hashed_backup_codes) and is_binary(backup_code) do
+    case Enum.find_index(hashed_backup_codes, &Crypto.verify_password(backup_code, &1)) do
+      nil ->
+        {:error, :invalid_backup_code}
+
+      index ->
+        {:ok, List.delete_at(hashed_backup_codes, index)}
```
sirn: I think I prefer this to be `used_at` and `invalidated_at` so they could be audit 🤔

sirn:
If my reading is right, it seems that `pot` can’t do a check for TOTP token “valid after” the last use (to prevent token reuse within interval). It does seems however that `pot:valid_totp` can validate a TOTP at timestamp (via Erlang datetime tuple) so I think we can do something like:

```
:pot.valid_totp(token, secret, [Timex.to_erl(user.last_logged_in)])
```

...to prevent token reuse. 🤔

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+    "object": "secret_code",
+    "label": "user@example.com",
+    "issuer": "OmiseGO"
+  }
+}
+```
+
+We can now create a two-factor QR code by using the data received in this response.
+
+> Try: using a web service to quickly generate the Two-factor QR Code e.g. [this](https://stefansundin.github.io/2fa-qr/)
```
unnawut:
```suggestion
**Try:** use a web service to quickly generate a two-factor QR code, e.g. [Stefan Sundin's 2FA QR code generator](https://stefansundin.github.io/2fa-qr/)
```

Let's keep quotes for actual quotes. Also link titles should be descriptive about the link for SEO purposes.

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+}
+```
+
+If everything was done correctly, the admin user account should now be secured with two-factor authentication.
+
+Next, we will look into how can we log in with 2FA.
+
+### Log in with 2FA
+
+After the admin user has enabled two-factor authentication successfully, the user will be need to do a 2-steps login to fully obtain the authentication token which we normally use to access the authenticated APIs.
```
unnawut:
```suggestion
After the admin user has enabled two-factor authentication successfully, the user will need to do a 2-step login to fully obtain the authentication token which we normally use to access the authenticated APIs.
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+-H "Accept: application/vnd.omisego.v1+json" \
+-H "Authorization: OMGAdmin $(echo -n "user_id:pre_auth_token" | base64 | tr -d '\n')" \
+-H "Content-Type: application/json" \
+-d '{
+  "passcode": "103405"
+}' \
+-v -w "\n" | jq
+```
+
+The response will be look similar to the previous login response, except the `pre_authentication_token`, which will be now a real `authentication_token`:
```
unnawut:
```suggestion
The response will look similar to the previous login response, except the `pre_authentication_token` which will now be the real `authentication_token`:
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+        "object": "authentication_token",
+        "master_admin": null,
+        "global_role": "super_admin",
+        "account_id": null,
+        "account": null
+    }
+}
+```
+
+Now, we can use the `user_id` and `authentication_token` from the response to form the valid authorization header, to be able to access the authenticated APIs.
```
unnawut:
```suggestion
Now, we can use the `user_id` and `authentication_token` from the response to form the valid authorization header, which can then be used to access the authenticated APIs.
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Disable 2FA
+
+To disable the two-factor authentication, calling to the api `/me.disable_2fa` as the following:
```
unnawut:
```suggestion
To disable the two-factor authentication, call the endpoint `/me.disable_2fa` as follows:
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+-H "Accept: application/vnd.omisego.v1+json" \
+-H "Authorization: OMGAdmin $(echo -n "user_id:auth_token" | base64 | tr -d '\n')" \
+-H "Content-Type: application/json" \
+-d '{
+  "passcode": "103405"
+}' \
+-v -w "\n" | jq
+```
+
+If the `passcode` is correct, the response will be look like as following:
```
unnawut:
```suggestion
If the `passcode` is correct, the response will look like the following:
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+{
+  "version": "1",
+  "success": true,
+  "data": {}
+}
+```
+
+### Note
+
+1. The user's authentication tokens are deleted after enabled or disabled two-factor authentication, so re-login might required.
```
unnawut:
```suggestion
1. The user's authentication tokens are deleted after two-factor authentication is enabled or disabled, so re-logins might be required.
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+
+2. `backup_code` can be used for disable and login.
```
unnawut:
```suggestion
2. `backup_code` can be used for login and/or disable two-factor authentication.
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+
+3. Each `backup_code` is one-time use.
```
unnawut:
```suggestion
3. Each `backup_code` can be used only once.
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+
+4. `pre_authentication_token` is returned only when the user has enabled two-factor authentication and log in with email and password, otherwise the  `authentication_token` is returned instead.
```
unnawut:
```suggestion
4. `pre_authentication_token` is returned only when the user has two-factor authentication enabled, otherwise the `authentication_token` is returned right after the user logs in with username and password.
```

docs/setup/advanced/settings.md
```diff
@@ -27,6 +27,8 @@ The eWallet is able to work independent of an external integration. The default
 
 - `min_password_length`: The minimum length of the password that a user is allowed to set.
 - `forget_password_request_lifetime`: The duration (in minutes) that a forget password request will be valid for.
+- `number_of_backup_codes`: The number of backup codes for the two-factor authentication.
+- `issuer`: The issuer for the two-factor authentication, which will be displayed the OTP app.
```
unnawut:
```suggestion
- `issuer`: The issuer of the two-factor authentication, which will be displayed on the TOTP app.
```

docs/guides/two_factor_authentication.md
```diff
@@ -0,0 +1,241 @@
...
+        "object": "pre_authentication_token",
+        "master_admin": null,
+        "global_role": "super_admin",
+        "account_id": null,
+        "account": null
+    }
+}
+```
+
+Second, calling to the two-factor authentication endpoint `/admin.login_2fa` by using the `pre_authentication_token` which we take from the previous step to form the authorization header, also specify the passcode which was generated by the TOTP authenticator app in the body like below:
```
unnawut:
```suggestion
Second, call the two-factor authentication endpoint `/admin.login_2fa` by using the `pre_authentication_token` from the previous step to form the authorization header. Also, specify the passcode generated by the TOTP authenticator app in the body like below:
```

apps/admin_api/test/admin_api/v1/controllers/two_factor_auth_controller_test.exs
```diff
@@ -0,0 +1,597 @@
...
+
+    test "responds error auth_token:not_found if continue using the old authentication token" do
+      user = User.get(@admin_id)
+      {%{backup_codes: [backup_code | _]}, _} = create_backup_codes(user)
+      {%{secret_2fa_code: secret_2fa_code}, _} = create_secret_code(user)
+      passcode = generate_totp(secret_2fa_code)
+
+      response =
+        admin_user_request("/me.enable_2fa", %{
+          "passcode" => passcode,
```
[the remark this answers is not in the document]
ripzery: Yep should accept `passcode` or `backup_code`. Forget to fix this test. The initial implementation required both.

apps/ewallet/lib/ewallet/auth/two_factor_authenticator.ex
```diff
@@ -0,0 +1,214 @@
...
+    - Update `enabled_2fa_at` to the current naive date time for user entity.
+    - Create a new auth_token. (The client must upgrade to this token)
+  - disable:
+    - Clear `enabled_2fa_at`, `hashed_backup_codes`, `secret_2fa_code` for user entity.
+  """
+
+  @default_number_of_backup_codes 10
+  @default_issuer "OmiseGO"
+
+  defmodule InvalidNumberOfBackupCodesError do
```
unnawut: 👍

apps/ewallet_config/config/config.exs
```diff
@@ -64,6 +64,21 @@ config :ewallet_config,
...
     },
+    "number_of_backup_codes" => %{
+      key: "number_of_backup_codes",
+      value: 10,
+      type: "unsigned_integer",
+      position: 106,
+      description: "The number of backup codes for the two-factor authentication."
+    },
+    "two_fa_issuer" => %{
+      key: "issuer",
```
unnawut:
```suggestion
      key: "two_fa_issuer",
```

apps/admin_api/test/admin_api/v1/controllers/two_factor_auth_controller_test.exs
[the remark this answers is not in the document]
unnawut: Can you test for `client:invalid_parameter` when both `passcode` and `backup_code` is given?

apps/admin_api/lib/admin_api/v1/controllers/two_factor_auth_controller.ex
```diff
@@ -0,0 +1,108 @@
...
+         {:ok, _} <- authorize(:create, conn.assigns, user),
+         {:ok, result} <- TwoFactorAuthenticator.create_and_update(user, type) do
+      render(conn, :create, result)
+    else
+      error -> respond_error(error, conn)
+    end
+  end
+
+  def enable(conn, %{"passcode" => _, "backup_code" => _}) do
+    respond_error({:error, :invalid_parameter}, conn)
```
unnawut:
```suggestion
    respond_error({:error, :invalid_parameter, "Invalid parameter provided. Either `passcode` or `backup_code` is required, but not both."}, conn)
```

unnawut: Actually this...

```elixir
def enable(conn, %{"backup_code" => _}) do
    respond_error({:error, :invalid_parameter, "Invalid parameter provided. `passcode` can be used to enable 2FA, but not `backup_code`}, conn)
end
```

ripzery:
> If my reading is right, it seems that `pot` can’t do a check for TOTP token “valid after” the last use (to prevent token reuse within interval). It does seems however that `pot:valid_totp` can validate a TOTP at timestamp (via Erlang datetime tuple) so I think we can do something like:
> 
> ```
> :pot.valid_totp(token, secret, [Timex.to_erl(user.last_logged_in)])
> ```
> 
> ...to prevent token reuse. 🤔

https://github.com/yuce/pot/commit/381d427bf1be6d37fa24f7a6bef856a1cc43bf4f

It seems that it only supports **create** a TOTP token for the given timestamp, but not for verify. 😢

apps/ewallet/lib/ewallet/auth/backup_code_authenticator.ex
```diff
@@ -55,20 +55,41 @@ defmodule EWallet.BackupCodeAuthenticator do
...
+       ) do
+    {:ok, hashed_backup_codes, [used_hashed_backup_code | used_hashed_backup_codes]}
+  end
+
+  defp handle_invalid_backup_code([], _), do: {:error, :invalid_backup_code}
+
+  defp handle_invalid_backup_code(used_hashed_backup_codes, backup_code) do
+    case Enum.any?(used_hashed_backup_codes, &Crypto.verify_password(backup_code, &1)) do
+      true ->
+        {:error, :used_backup_code}
```
sirn: I think we shouldn't expose the fact that the backup code was already used. Maybe just return `:invalid_backup_code` for all cases?

apps/ewallet_db/priv/repo/migrations/20190510071247_add_two_fa_to_user.exs
```diff
@@ -0,0 +1,13 @@
+defmodule EWalletDB.Repo.Migrations.AddTwoFAToUser do
+  use Ecto.Migration
+
+  def change do
+    alter table(:user) do
+      add :enabled_2fa_at, :naive_datetime
+      add :secret_2fa_code, :string
+      add(:hashed_backup_codes, {:array, :string}, default: [])
+      add :used_backup_code_at, :naive_datetime
+      add(:used_hashed_backup_codes, {:array, :string}, default: [])
```
sirn: PostgreSQL array can get slow really fast, and the way to manipulate them by _updating_ the field may subject to race condition without proper locking (e.g. when user try to use two backup codes at the same time from different session).

I think it's better to follow [1NF](https://en.wikipedia.org/wiki/First_normal_form) design and have a `user_backup_code` table that store an ID of a user, a hashed backup code, a date of creation, and a date of use (in which setting date of use invalidated the code). 

Sorry, I just caught this.

ripzery: 👍 Will fix it

ripzery: @sirn Done krub
Implementation: https://github.com/omisego/ewallet/pull/920/commits/926849a130f5a420496d042ffe97576eeed83673
Fixed tests and add some:
https://github.com/omisego/ewallet/pull/920/commits/8ce8b385621fd88dc1104cf67a7d433227ec55c6

ripzery: @sirn Please re-review 🙏 

https://github.com/omisego/ewallet/pull/920/files/894af2be4206e0e8d68d9469c8a368196a530e7f..52e97a72157c13f8bc0ff0ad988f229e2b48ef66

apps/ewallet_db/lib/ewallet_db/user_backup_code.ex
```diff
@@ -0,0 +1,139 @@
...
+    |> build_multiple_attrs()
+    |> map_changeset()
+    |> do_insert()
+  end
+
+  defp do_insert(changesets) when is_list(changesets) do
+    changesets
+    |> Enum.with_index()
+    |> Enum.reduce(Multi.new(), fn {changeset, index}, acc ->
+      Multi.insert(acc, :"ubc_#{index}", changeset)
```
unnawut: `ubc_#{index}` is quite hard to guess what it stands for. Can we use its full name of the action? e.g. `insert_user_backup_code_#{index}`?

Issue #895, opened 2019-03-27
Add master_account setting in the configuration page


T-Dnzt:
A new setting named `master_account` was added and needs to be shown on the configuration page. The value that can be set is an account ID.

PR #929, opened 2019-04-04
masteraccount setting on configurations page

nicholasmueller:
Issue/Task Number: [895](https://github.com/omisego/ewallet/issues/895)

# Overview

A new setting named master_account was added and needs to be shown on the configuration page. The value that can be set is an account ID.

# Usage

Change master account setting in config, save, and make sure its there on reload.
Also check accounts with and without avatar images


apps/admin_panel/assets/src/omg-account-select/index.js
```diff
@@ -0,0 +1,41 @@
...
+  color: ${props => props.theme.colors.B100};
+  font-size: 10px;
+`
+const AccountSelect = ({ account }) => {
+	const name = _.get(account, 'name')
+	const identifier = _.get(account, 'id')
+	const thumbnail = _.get(account, 'avatar.thumb');
+
+	return (
+		<AccountSelectContainer>
```
nicholasmueller: @jarindr  🤕 whyyyyy is the spacing like this... my config on vscode is 2 spaces

jarindr: which formatter you’re using tho

nicholasmueller: @jarindr actually i'm not using any formatter 🤔 
maybe because when i created the file it defaulted to 4 spaces, then i changed it, so github still shows it this way? can you see in your code editor if its correctly showing 2 spaces... maybe its just github?

T-Dnzt:
Looking good, just a few comments:

- autocomplete doesn't seem to work (it should filter by name / ID automatically)
- Can we move it to the top of the list? Above Base URL
- Can we add the name of the selected account under or next to the field?
- The UX when sending an invalid value (like `123`) seems to work, but doesn't actually send that value (which would return an error from the server). Maybe we shouldn't let users Save if they haven't picked an account?

PR #930, opened 2019-04-04
Handle client reset password

jarindr:
Issue/Task Number: #908 

# Overview

Handle reset password on client.

# Changes

- Add reset password fallback to app waiting scene.

# Implementation Details

There are many possibility for this,
- if it is a mobile, and foward_url exist we try to open an app and give a scene to wait.
- if not show usual form.

# Usage

try on mobile, and also opening an app.

# Impact

Deploy as usual.

### DIFF
--- a/apps/admin_panel/assets/src/clientApp/omg-client-reset-password/index.js
+++ b/apps/admin_panel/assets/src/clientApp/omg-client-reset-password/index.js
@@ -7,7 +7,6 @@ import PropTypes from 'prop-types'
 import AuthFormLayout from '../../omg-layout/authFormLayout'
 import { compose } from 'recompose'
 import { updateUserPassword } from '../../services/clientService'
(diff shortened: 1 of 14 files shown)


apps/admin_panel/assets/src/clientApp/omg-client-reset-password/index.js
```diff
@@ -0,0 +1,176 @@
...
+      const result = await updateUserPassword({
+        email,
+        token,
+        password: this.state.newPassword,
+        passwordConfirmation: this.state.reEnteredNewPassword
+      })
+      if (result.data.success) {
+        this.setState({ submitStatus: 'SUCCESS' })
+      } else {
+        this.setState({ submitStatus: 'FAILED', submitErrorText: result.data.data.code })
```
mederic-p: Can we use the description instead?

Issue #383, opened 2018-07-26
Allow existing groups to have new categories in Admin Panel


T-Dnzt:
Currently, categories can only be set when creating a new account. We support the updating of accounts with categories, this should be reflected in the admin panel by showing a field to add categories to existing accounts.

PR #935, opened 2019-04-09
Update categories in account settings

nicholasmueller:
Issue/Task Number: [383](https://github.com/omisego/ewallet/issues/383)

# Overview

Adds a select category field to account settings

# Changes

send the selected category with the update account call

# Implementation Details

- "Create account" currently only supports selecting one category, so i did the same here
- Allows user to save no category by emptying field
- Dont allow user to click save unless category is selected from dropdown


T-Dnzt:
Can you also allow the creation of a category on the account settings page?

Issue #683, opened 2019-01-21
Handle secondary wallets for accounts in Admin Panel


T-Dnzt:
Currently, the Admin Panel only supports showing and creating primary wallets for accounts. The API supports the creation of secondary wallets for accounts, and this ability should be reflected in the Admin Panel.

PR #939, opened 2019-04-18
Add secondary wallets

nicholasmueller:
Issue/Task Number: [683](https://github.com/omisego/ewallet/issues/683)

# Overview

Be able to create secondary/burn wallets from the account wallet page in the admin panel

# Changes

- New create wallet modal


T-Dnzt:
Looking good, but can you remove the `Create` button from the global Wallets page? it should only be shown under Account > Wallets (else it will create a wallet for I don't know which account).

PR #944, opened 2019-04-19
Frontend lint & upgrade packages

nicholasmueller:
Ran the linter on the frontend and made it happy 

also upgraded everything. result of that was only to fix some styled components upgrade warnings, specifically with the usage of `innerRef` and `attrs`

also fixed some other warnings like invalid proptypes, and invalid dom attributes from forwarding all `{...props}`where we probably shouldn't

please reinstall packages, and go through the app to make sure i didnt break anything since this touches **_a lot_** of files


sirn:
I would like to integrate this to the `lint` step in Circle to prevent @jarindr from breaking lint (:trollface:), what command do I need to run?

nicholasmueller:
@sirn `npm run lint`

Issue #965, opened 2019-04-24
Fix 1.2 design based on QC report


T-Dnzt:

unnawut:
Closed by #976

PR #976, opened 2019-04-30
QC 1.2

nicholasmueller:
Issue/Task Number: [965](https://github.com/omisego/ewallet/issues/965)

To review please refer to all the comments in this series -> https://gallery.io/projects/MCHbtQVoQ2HCZfyN0ITInFtb/files/MCEJu8Y2hyDScV4P233AvoydBkws7ZzgNM0


apps/admin_panel/assets/src/omg-page-activity-log/index.js
```diff
@@ -243,6 +257,7 @@ class AccountPage extends Component {
           loadingStatus={individualLoadingStatus}
         >
           <SortableTable
+            activeIndexKey={this.state.selectedRow}
```
jarindr: const activeIndexKey = queryString.parse(this.props.location.search)['show-activity-tab']
then you can remove all the state

jarindr:
Screen Shot 2019-04-30 at 2 00 23 PM

And this as we discussed, I think we should either just use a text or remove border when disabled

PR #979, opened 2019-05-02
Geth ethereum node communication

### LINKED ISSUE #693: Ethereum Node Communication (with Geth adapter)
Our original implementation will only support Geth, but it should be done in a way that allows the switch to other types of nodes later on. Therefore, adapter modules needs to be implemented, starting with the Geth one.

We should only rely on the `JSON-RPC` API, because, while the eWallet will ship with an Ethereum node (and could potentially use the IPC API), it is actually possible to connect to remote nodes as well (as this will be the case at first).

 __No Ethereum accounts__ should be generated/stored in Geth.

mederic-p:
Issue/Task Number: #693 
Closes #693 

# Overview

This PR refactors the `blockchain` and `eth_blockchain` sub-apps and adds the support for the `geth` adapter.

# Changes

- Rename `blockchain` to `eth_blockchain`
- Rename `blockchain_eth` to `geth_eth_adapter`
- Move wallet generation to `eth_blockchain` sub-app
- Add `ethereumex` and `abi` dependencies to handle JSON-RPC calls to `geth`
- Add genserver handler for `{:get_balances, address, contract_addresses, block}` that returns a map of contract_address -> balance for the specified wallet address
- Add `default_adapter` config that is set to `:geth` by default except for tests where it's set to `:dumb` which is a `Dumb` adapter that mock responses.

# Usage

- Set the url of a JSON-RPC server in `eth_geth_adapter/config/dev.exs`
- Have a valid wallet address with an eth balance and optionally a few ERC20 tokens balances.
- Run the elixir console: `iex -S mix`
- Call 
```
EthBlockchain.Balance.get({"0x54e0588607dcec6c0b36fca1154a57814a913591",  ["0x48b91d5f363892592bf836777dc73b54a10b72ae", "0x0000000000000000000000000000000000000000", "0x436dbc182828ccd421238fb622613c953c5e5028", "0xb76fe945f8672b57cc151448e2f52b0f88891c34"]})
```
(replace the first param with your wallet address and the list of addresses with the actual contract addresses).
> Note: the `0x0000000000000000000000000000000000000000` represents the `Eth` token.
- Receive a response:
```
{:ok,
 %{
   "0x0000000000000000000000000000000000000000" => 13155000000000000000000,
   "0x436dbc182828ccd421238fb622613c953c5e5028" => 10000000000,
   "0x48b91d5f363892592bf836777dc73b54a10b72ae" => 999999999999999999999999900,
   "0xb76fe945f8672b57cc151448e2f52b0f88891c34" => 0 
 }}
```

# For future PRs:

- Refactor response handler/parser if new JSON call need to use them
- Improve the setting/config setup for the blockchain adapters


apps/eth_blockchain/test/eth_blockchain/adapter_test.exs
```diff
@@ -0,0 +1,149 @@
...
+
+defmodule EthBlockchain.AdapterTest do
+  use ExUnit.Case
+  alias EthBlockchain.{Adapter, DumbAdapter}
+  alias Ecto.UUID
+
+  setup do
+    mock_balance = 123
+
+    Code.eval_string("""
```
sirn: Should we remove this and use only `dumb` adapter?

apps/eth_blockchain/test/eth_blockchain/adapter_test.exs
```diff
@@ -0,0 +1,149 @@
...
+        def handle_call({:get_balances, _address, contract_addresses, _block}, _from, reg) do
+          balances = Map.new(contract_addresses, fn ca -> {ca, #{mock_balance}} end)
+          {:reply, {:ok, balances}, reg}
+        end
+      end
+    """)
+
+    supervisor = String.to_atom("#{UUID.generate()}")
+
+    {:ok, _} =
```
sirn: Might worth switching this to `EthBlockchainCase`. Probably make it return supervisor pid as well? (Since we need to perform the `which_children` check?)

apps/eth_geth_adapter/config/prod.exs
```diff
@@ -0,0 +1,3 @@
+use Mix.Config
+
+config :ethereumex, url: "http://localhost:8545", client_type: :http
```
sirn: Maybe we should ask Mana-Ethereum people to support `{:system, ...}` style config 😉

apps/eth_geth_adapter/test/eth_geth_adapter/worker_test.exs
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule EthGethAdapter.WorkerTest do
+  use ExUnit.Case
+  alias Ecto.Adapters.SQL.Sandbox
+  alias Keychain.Repo
+
+  setup tags do
+    :ok = Sandbox.checkout(Repo)
```
sirn: I think the worker shouldn’t access `Keychain` directly. Instead, let the `EthBlockchain` pass it to the adapter (so the adapter can be self-contained).

docs/design/components.md
```diff
@@ -26,6 +26,6 @@ The eWallet is an umbrella Elixir application containing the following sub-appli
 
   - [load_tester](/apps/load_tester): Sub-application containing the load test runner and its scripts.
 
-## Sub-applications planned
+  - [eth_blockchain](/apps/eth_blockchain) Interface to the Ethereum blockchain and its adapters
```
unnawut: 👏👏👏 for remembering to update this doc

apps/eth_geth_adapter/lib/eth_geth_adapter/balance.ex
```diff
@@ -0,0 +1,113 @@
...
+  ```
+  if successful or {:error, error_code} if failed.
+  """
+  def get(address, contract_address, block \\ "latest")
+
+  def get(address, contract_addresses, block) when is_list(contract_addresses) do
+    case ERC20.abi_balance_of(address) do
+      {:ok, encoded_abi_data} ->
+        contract_addresses
+        |> Enum.reduce([], fn contract_address, acc ->
```
unnawut: [`Enum.map/2`](https://hexdocs.pm/elixir/Enum.html#map/2) might give a little cleaner code also without need to reverse?

PR #981, opened 2019-05-03
Add blockchain wallet schema

### LINKED ISSUE #690: Blockchain wallet schema

mederic-p:
Issue/Task Number:  #690 
Closes #690 

# Overview

This PR adds `blockchain_address` to `Token` and introduce the new `BlockchainWallet` entity. 

# Changes

- Add `blockchain_address` to `Token`
- Add `BlockchainWallet`


apps/ewallet_db/test/ewallet_db/token_test.exs
```diff
@@ -99,12 +100,13 @@ defmodule EWalletDB.TokenTest do
   end
 
   describe "update/2" do
-    test_insert_field_length(Token, :iso_code)
-    test_insert_field_length(Token, :name)
-    test_insert_field_length(Token, :description)
-    test_insert_field_length(Token, :short_symbol)
-    test_insert_field_length(Token, :html_entity)
-    test_insert_field_length(Token, :iso_numeric)
+    test_update_field_length(Token, :iso_code)
```
unnawut: 😅 🙏

apps/ewallet_db/test/ewallet_db/token_test.exs
```diff
@@ -166,6 +168,62 @@ defmodule EWalletDB.TokenTest do
...
+    test "returns a query of tokens that have a blockchain address" do
+      assert Enum.empty?(Token.all())
+
+      :token |> params_for(%{blockchain_address: "0x01"}) |> Token.insert()
+      :token |> params_for(%{blockchain_address: "0x02"}) |> Token.insert()
+      :token |> params_for() |> Token.insert()
+
+      tokens = Token.query_all_blockchain() |> Repo.all()
+
+      assert length(tokens) == 2
```
unnawut: Can you assert blockchain address too? Just in case the 2 happen to be non blockchain.

apps/ewallet_db/lib/ewallet_db/blockchain_wallet.ex
```diff
@@ -0,0 +1,85 @@
...
+    |> cast_and_validate_required_for_activity_log(
+      attrs,
+      cast: [
+        :uuid,
+        :address,
+        :public_key,
+        :name,
+        :type
+      ],
+      required: [:address, :name, :public_key, :type]
```
unnawut: Should we require `:name` if this `BlockchainWallet` will also be user's wallets in the future?

mederic-p: So, after discussing with Tibo, we should keep it required for now and potentially remove it later if needed.

Issue #980, opened 2019-05-02
Transfer modal exchange rate is borken with x,xxx type of amunt


jarindr:

unnawut:
Closed with #984

PR #984, opened 2019-05-03
Fix rate exchange bug

jarindr:
Issue/Task Number: #980 

# Overview

When creating rate, the amount is not passed correctly because it divided a string `x,xxx` / `y,yyy`.

# Changes

- Fix rate dividing.
- Fix height and unused css for top navigation.
- Fix edit rate returning [Object]
# Impact

Deploy as usual

### DIFF
--- a/apps/admin_panel/assets/src/omg-exchange-rate-modal/createExchangeModal.js
+++ b/apps/admin_panel/assets/src/omg-exchange-rate-modal/createExchangeModal.js
@@ -318,7 +318,7 @@ class CreateExchangeRateModal extends Component {
                       value={this.state.fromTokenRate}
                       onChange={this.onChangeRate('fromToken')}
                       type='amount'
-                      maxLength={18}
+                      maxAmountLength={18}
                       normalPlaceholder={0}
                       suffix={this.state.fromTokenSymbol}
                     />
@@ -359,7 +359,7 @@ class CreateExchangeRateModal extends Component {
                       value={this.state.toTokenRate}
                       onChange={this.onChangeRate('toToken')}
                       type='amount'
-                      maxLength={18}
+                      maxAmountLength={18}
                       step='any'
                       normalPlaceholder={0}
                       suffix={this.state.toTokenSymbol}
--- a/apps/admin_panel/assets/src/omg-exchange-rate-modal/createExchangeModal.js
+++ b/apps/admin_panel/assets/src/omg-exchange-rate-modal/createExchangeModal.js
@@ -315,6 +315,7 @@ class CreateExchangeRateModal extends Component {
                       value={this.state.fromTokenRate}
                       onChange={this.onChangeRate('fromToken')}
                       type='amount'
+                      maxLength={18}
                       normalPlaceholder={0}
                       suffix={this.state.fromTokenSymbol}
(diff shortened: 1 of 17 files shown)


apps/admin_panel/assets/src/omg-exchange-rate-modal/createExchangeModal.js
```diff
@@ -231,30 +235,25 @@ class CreateExchangeRateModal extends Component {
       onlyOneWayExchange,
       oppositeExchangePair
     } = this.state
-
-    const forwardRate = _.round(toTokenRate / fromTokenRate, 3)
-    const backRate = _.round(1 / forwardRate, 3)
-
+    const forwardRate = +new BigNumber(numeral(toTokenRate).value()).dividedBy(numeral(fromTokenRate).value()).toFixed(8)
+    const backRate = new BigNumber(1).dividedBy(new BigNumber(numeral(forwardRate).value())).toFixed(8)
+    console.log(backRate)
```
nicholasmueller: log!

apps/admin_panel/assets/src/omg-exchange-rate-modal/createExchangeModal.js
```diff
@@ -231,30 +235,25 @@ class CreateExchangeRateModal extends Component {
       onlyOneWayExchange,
       oppositeExchangePair
     } = this.state
-
-    const forwardRate = _.round(toTokenRate / fromTokenRate, 3)
-    const backRate = _.round(1 / forwardRate, 3)
-
+    const forwardRate = +new BigNumber(numeral(toTokenRate).value()).dividedBy(numeral(fromTokenRate).value()).toFixed(8)
+    const backRate = new BigNumber(1).dividedBy(new BigNumber(numeral(forwardRate).value())).toFixed(8)
```
unnawut: `backRate` is not necessarily the inverse of `forwardRate` though, that's why we have separate pairs on the backend to support exchange rates that usually have a spread.

nicholasmueller: Screen Shot 2562-05-07 at 11 02 05 (2)

i guess backRate now is just the suggested(?) opposite rate, with zero spread... maybe we should add a 'spread' input?

apps/admin_panel/assets/src/omg-exchange-rate-modal/createExchangeModal.js
unnawut: 😱

unnawut: The linter doesn't complain about this?

nicholasmueller:
also edit modal `updatePair` button broken.. i think because of the commas in the input...

apps/admin_panel/assets/src/omg-exchange-rate-modal/createExchangeModal.js
unnawut: Ah since it's only a suggestion and can be edited without changing the forward rate, this should be fine.

jarindr:
@nicholasmueller fixed both problem

PR #1018, opened 2019-05-23
Refactor modal and user page

jarindr:
Issue/Task Number: #252 

# Overview

This PR includes many things when doing user page, because I need to reuse many parts of the app So I think it is a good time to do a big refactoring.

# Changes

- Add new User detail page.
- Refactor table for transaction, activity. Since we need to reuse in many pages.
- Refactor out Create transaction modal so you can open the modal with action.
- Add env example
# Implementation Details

- Move modal out from components, open with action.

# Usage

- Click on user table row, see the new user detail.

# Impact

Deploy as usual after fixing conflicts.


apps/frontend/assets/src/omg-modal/ModalController.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import { selectGetModalById } from './selector'
+import { closeModal, openModal } from './action'
+
+import CreateTransactionModal from '../omg-create-transaction-modal'
+import _ from 'lodash'
+
+// ADD YOUR NEW MODAL HERE
```
T-Dnzt: 🕺

PR #1045, opened 2019-06-10
Remove yarn

jarindr:
Issue/Task Number: -

# Overview

Remove yarn installation and use default npm.

# Changes

- Remove yarn installation and use default npm.

# Implementation Details

-

# Usage

-

# Impact

Not sure, need @sirn to comfirm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ deps-ewallet:
 	mix deps.get
 
 deps-assets:
-	$(ASSETS) yarn install
+	$(ASSETS) npm install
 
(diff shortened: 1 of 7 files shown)


sirn:
Why reverting back to npm?

jarindr:
@sirn yarn was much better, but now npm catches up with the speed and reliability that yarn provided in 2017-2018

So I think we should switch back to npm, we can also remove yarn deps when installing the ewallet.

related article that think the same as me https://medium.com/@vincentnewkirk/npm-vs-yarn-2019-e88757b17038

PR #1054, opened 2019-06-12
Minor fix transaction consumption confirmation

### LINKED ISSUE #1053: Allow to confirm a transaction consumption regardless of transaction request expiration
Currently, it fails and returns the expiration reason.

Refer to the following scenario:
```
Noting this discussion down for reference. Consider this scenario:

1. Provider creates a transaction request
2. User consumes the transaction request with amount override -> pending consumption created
3. Provider cancelled the transaction request created in Step 1
4. Provider approves the consumption from Step 2 -> a successful transaction occurs

This is allowed since there can be cases where a provider wishes to stop accepting new payments but allow users that are mid-way through the process to continue to the end.
```

_Originally posted by @unnawut in https://github.com/omisego/ewallet/pull/1046#issuecomment-500714144_

ripzery:
Issue/Task Number: #1053 

Closes #1053 

# Overview

This PR allows the transaction consumption with an expired transaction request with `expiration_reason = "expired_transaction_request"` to be confirmed

> ⚠️  Waiting branch 950-add-endpoint-cancel-transaction-request to be merged

### DIFF
--- a/apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
+++ b/apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
@@ -118,6 +118,13 @@ defmodule EWallet.TransactionConsumptionValidator do
   end
 
   defp validate_transaction_request(request, operation) do
+<<<<<<< HEAD
+=======
+    expiration_reason =
+      request.expiration_reason &&
+        String.to_existing_atom(request.expiration_reason)
+
+>>>>>>> f0844ca37... :hammer: Refactor handling `expiration_reason`
     request
     |> TransactionRequest.valid?()
     |> Kernel.||(TransactionRequest.get_expiration_reason(request))
--- a/apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
+++ b/apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
@@ -146,11 +146,11 @@ defmodule EWallet.TransactionConsumptionValidator do
       true ->
         :ok
 
-      :cancelled_transaction_request ->
-        :ok
+      :max_consumptions_reached ->
+        {:error, :max_consumptions_reached}
 
       expiration_reason when not is_nil(expiration_reason) and is_atom(expiration_reason) ->
-        {:error, expiration_reason}
+        :ok
 
       _ ->
         {:error, :unknown_error}
--- a/apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
+++ b/apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
@@ -118,9 +118,7 @@ defmodule EWallet.TransactionConsumptionValidator do
   end
 
   defp validate_transaction_request(request, operation) do
-    expiration_reason =
-      request.expiration_reason &&
-        String.to_existing_atom(request.expiration_reason)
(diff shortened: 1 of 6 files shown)


apps/ewallet/lib/ewallet/validators/transaction_consumption_validator.ex
```diff
@@ -149,6 +149,9 @@ defmodule EWallet.TransactionConsumptionValidator do
       :cancelled_transaction_request ->
         :ok
 
+      "expired_transaction_request" ->
```
unnawut: How about applying `String.to_existing_atom/1` since `|> Kernel.||(request.expiration_reason)`? 
This way we have more consistent typing + work is done closer to the source.

T-Dnzt: I like the idea.

ripzery: The `request.expiration_reason` can be nil, so need to validate it before converting to :atom.

In my thought, it will be become this.

```elixir
expiration_reason =
      request.expiration_reason &&
        String.to_existing_atom(request.expiration_reason)

    request
    |> TransactionRequest.valid?()
    |> Kernel.||(expiration_reason)
    |> Kernel.||(TransactionRequest.get_cancelled_error(request))
    |> do_validate_transaction_request(operation)
```
Not sure if I have a better choice 🤔

unnawut: Looks fine to me, although I'd like it explicit than a boolean shortcut e.g.

```elixir
expiration_reason =
  case request.expiration_reason do
    nil -> nil
    reason -> String.to_existing_atom(reason)
  end
```

Could turn this into its own function in `TransactionRequest` as well, can imagine it being re-used in other places.

ripzery: oh right would be better to move to `TransactionRequest` 🙇

apps/ewallet/test/ewallet/gates/transaction_consumption_confirmer_gate_test.exs
```diff
@@ -708,7 +709,8 @@ defmodule EWallet.TransactionConsumptionConfirmerGateTest do
           %System{}
         )
 
-      assert res == {:error, :expired_transaction_request}
+      assert status == :ok
```
unnawut: But the test name is `"fails to confirm the consumption if expired"`?

apps/ewallet/test/ewallet/validators/transaction_consumption_validator_test.exs
```diff
@@ -179,11 +179,40 @@ defmodule EWallet.TransactionConsumptionValidatorTest do
           end_user: user
         })
 
-      assert status == :error
-      assert res == :expired_transaction_request
+      assert status == :ok
```
unnawut: But the test name is `"expires request if past expiration date"`?

apps/ewallet/test/ewallet/gates/transaction_consumption_confirmer_gate_test.exs
```diff
@@ -379,7 +379,8 @@ defmodule EWallet.TransactionConsumptionConfirmerGateTest do
...
```
unnawut: But the test name is for `"fails to confirm the consumption if expired"`?

Issue #715, opened 2019-01-22
Advanced filtering for transactions


T-Dnzt:

unnawut:
Examples?

unnawut:
My try:
- Filter between local ledger / blockchain

PR #1064, opened 2019-06-20
Advanced Filtering

nicholasmueller:
Issue/Task Number: [715 ](https://github.com/omisego/ewallet/issues/715)

# Overview

depends on https://github.com/omisego/ewallet/pull/1066

Advanced filtering as per this design series -> https://gallery.io/projects/MCHbtQVoQ2HCZfyN0ITInFtb/files/MCHbtQVoQ2HCZeFc8VHDAU1d

Also some filters as per the design are not possible so weren't included

# Changes

- configurable filters in one place (`FilterMap.js`) so you can add, remove, tweak in the future
- new react-select dependency for multi select dropdown

# Usage

Requires a lot of testing, each filter on each page as well as the combination of filters.
Navigate pages, add filters, remove filters, remove breadcrumbs, try different use cases.

how to use on other pages -> add <AdvancedFilter> wherever you want the filter crumbs to show and it will pass a `{ matchAll, matchAny }` object to the `onFilter` prop when filters are defined. You can then inject that to the fetcher query on that page (using local state in this case).


mederic-p:
Overall:
    - Create Date -> Creation date? Also I’d change `Start` `End` to `From` To`
    - I’d prefer to see the filters as a popup that shows under the `Filter` button when you click on it instead of a plain page but need to talk with the designer first
    - The `Yes` `No` choices could be replaced with a tick box (to discuss with designer)
    - Would be nice to have an autocompleted list showing when we start typing on some fields (wallet/user/…)
    - Can we somehow “store” the filter infos in the url (so we can share a link with all filtering info)
	

Transaction filters: 
    - transfer to / specify target don’t list all possible accounts (missing user accounts)
Transaction Request filters: 
    - Received -> should be Receive
    - Confirmation Type -> Should be Requires confirmation
    - Specify Requested Wallet -> Specify Request Wallet
    - Can we use the same dropdown picker as in transactions (the one with more details about accounts)
    - Would be nice to be able to filter by token and amount also
Transaction Consumption filters: 
    - Same as Transaction Request
    - Consumed Date -> Consumption Date

jarindr:
i'm ok on the implementation wise, for the design improvement can deal with it later I think.

apps/frontend/assets/src/adminPanelApp/globalStyle.css
```diff
@@ -37,7 +37,7 @@ body {
   top: 0;
   right: 0;
   bottom: 0;
-  overflow: hidden;
+  overflow-y: scroll;
```
jarindr: the body overflow should always be hidden, the overflow should be handle on modal side.

jarindr:
Update:
- the filter of transfer from/to when typing doesn't change the dropdown selection items.
- Specify Target also doesn't work.

nicholasmueller:
@mederic-p can you please give it another look? i made all the selects searchable and removed the multi-selects (annoying bug in the dependency I am using where you can't dynamically update the dropdown list)

apps/frontend/assets/src/omg-uikit/timepicker/index.js
```diff
@@ -0,0 +1,48 @@
+import React, { PureComponent } from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import DateTime from 'react-datetime'
+
+import { Input } from '..'
```
jarindr: no `..` please 😆

apps/frontend/assets/src/omg-uikit/datepicker/index.js
```diff
@@ -0,0 +1,48 @@
+import React, { PureComponent } from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import DateTime from 'react-datetime'
+
+import { Input } from '..'
```
jarindr: here too

PR #1098, opened 2019-07-05
new select wrapper

nicholasmueller:
issue #1102

ui component for new design. wrapper over select & input.
Implemented it for global transfer modal. 

https://gallery.io/projects/MCHbtQVoQ2HCZfyN0ITInFtb/files/MCEJu8Y2hyDScUKnMG1qETPXk0_H_OMFTOk

### DIFF
--- a/apps/frontend/assets/src/omg-create-transaction-modal/index.js
+++ b/apps/frontend/assets/src/omg-create-transaction-modal/index.js
@@ -5,7 +5,7 @@ import { connect } from 'react-redux'
 import { compose } from 'recompose'
 import { withRouter } from 'react-router-dom'
 
-import { Input, Button, Icon, Select, SelectInput } from '../omg-uikit'
+import { Button, Icon, SelectInput } from '../omg-uikit'
 import Modal from '../omg-modal'
 import { transfer } from '../omg-transaction/action'
 import { getWalletById } from '../omg-wallet/action'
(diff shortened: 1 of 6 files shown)


jarindr:
when typing something else that is not number on amount, it breaks

apps/frontend/assets/src/omg-create-transaction-modal/index.js
```diff
@@ -251,68 +239,83 @@ class CreateTransaction extends Component {
...
+                selectProps={{
+                  label: 'Wallet Address',
+                  clearable: true,
+                  disabled: !!this.props.fromAddress,
+                  onSelectItem: this.onSelectFromAddressSelect,
+                  value: this.state.fromAddress,
+                  onChange: this.onChangeInputFromAddress,
+                  valueRenderer: this.state.fromAddressSelect
+                    ? value => {
+                      const wallet = _.find(data, i => i.address === value)
```
jarindr: why do we need to filter to find address again ? since we already search it from the api

nicholasmueller: because the `value` available to the component is not the entire object the `valueRenderer` needs. would need to refactor the select to store the object `onSelect` if we want to keep it locally

nicholasmueller:
@jarindr fixed the input non number issue

PR #1111, opened 2019-07-10
Create cold wallet references

### LINKED ISSUE #695: Update Token and add ETH generating module
Requires #693

The ETH token should automatically be added when the blockchain adapter is enabled and set to Ethereum. 

1. First, we need to give the capacity to our internal tokens to become blockchain tokens and contains blockchain details (such as the smart contract address for ERC-20 tokens). 
2. Each blockchain adapter will need a booting function to prepare the eWallet to support that blockchain. In Ethereum's case, this would involve generating the ETH token.

- [ ] Turn internal tokens into blockchain tokens.
- [ ] Add setup function to the Ethereum adapter we've built.
- [ ] Generate the ETH token if not already present.

mederic-p:
Issue/Task Number: 695
Closes #695, #1107 

# Overview

This PR allows the creation of a cold wallet "reference" using `blockchain_wallet.create`.

# Changes

- Add `blockchain_wallet.create` endpoint and associated files.

# Implementation Details

The `blockchain_wallet.create` endpoints requires a `name`, an `address` and a `type`. For now, as we are only supporting cold wallet creation with the API, the given `type` must be `cold`.

# Usage

The easiest is to test the endpoint in the updated swagger ui.

# Impact

N/A


apps/ewallet_db/priv/repo/migrations/20190710085358_add_blockchain_identifier_to_blockchain_wallet.exs
```diff
@@ -0,0 +1,22 @@
...
+  alias EWalletDB.Repo
+
+  def change do
+    alter table(:blockchain_wallet) do
+      add :blockchain_identifier, :string
+    end
+
+    flush()
+
+    {:ok, _} = Repo.query("UPDATE blockchain_wallet SET blockchain_identifier = 'ethereum' WHERE blockchain_identifier IS NULL")
```
unnawut: Does this work for both up and down migrations? 🤔

apps/ewallet_db/priv/repo/migrations/20190710085358_add_blockchain_identifier_to_blockchain_wallet.exs
```diff
@@ -0,0 +1,22 @@
...
+
+    drop constraint(:transaction, "transaction_from_blockchain_address_fkey")
```
unnawut: Also this one, not sure it works for down.

apps/ewallet_db/lib/ewallet_db/blockchain_wallet.ex
```diff
@@ -103,7 +134,27 @@ defmodule EWalletDB.BlockchainWallet do
   @spec insert(map()) :: {:ok, %__MODULE__{}}
   def insert(attrs) do
```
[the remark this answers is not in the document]
unnawut: Noted that it's being used by the test helpers. Refactor can be done in a separate PR

PR #1135, opened 2019-08-13
Add nonce handler

### LINKED ISSUE #1088: Add nonce GenServer to allow transaction chaining.

mederic-p:
Issue/Task Number: 1088
Closes #1088 

# Overview

This PR adds a nonce handler that will give an incremented nonce for each new transaction from the same account. This allows an address to do more than 1 transaction per block.

# Changes

- Add `EthBlockchain.Nonce` module
- Update `EthBlockchain.Transaction` to take nonce from the new module

### DIFF
--- a/apps/eth_blockchain/lib/eth_blockchain/nonce.ex
+++ b/apps/eth_blockchain/lib/eth_blockchain/nonce.ex
@@ -1,0 +1,105 @@
+# Copyright 2018-2019 OmiseGO Pte Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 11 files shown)


apps/eth_blockchain/lib/eth_blockchain/nonce.ex
```diff
@@ -0,0 +1,105 @@
...
+    name = Keyword.get(opts, :name, __MODULE__)
+    GenServer.start_link(__MODULE__, %{}, name: name)
+  end
+
+  def init(_args) do
+    {:ok, %{}}
+  end
+
+  def handle_call({:next_nonce, address, node_adapter, blockchain_adapter_pid}, _from, state) do
+    case Map.get(state, address) do
```
unnawut: Might be a bottleneck/single point of failure to have all nonce data stored in one GenServer/process.  

Maybe 1 GenServer per address, or use [`ets`](https://elixir-lang.org/getting-started/mix-otp/ets.html) instead without GenServer?

PR #1147, opened 2019-09-03
Deposit wallet funds pooling

### LINKED ISSUE #1144: Move funds to hot wallet after deposit

unnawut:
Issue/Task Number: #1144 
Closes #1144 

# Overview

Deposit wallet balance management by automatically pooling funds from multiple deposit wallets into a hot wallet.

# Changes

See implementation details.

# Implementation Details

## Application config and startup
- Ethereum's chain_id now configurable through `ROOTCHAIN_CHAIN_ID` env var
- `EWallet.Application`'s child specs updated away from deprecated `worker/2` `supervisor/3` to tuple-based specs
- New application config: `deposit_pooling_interval` (default to 1 hour, disabled in test environment by setting to `0`)

## Transaction creation
- TransactionTracker's API e.g. lookup, create, etc. all go through `TransactionTracker` instead of both `TransactionTracker` and `TransactionRegistry`, etc.
- Updated `EthBlockchain.Transaction`
  - `create_contract/1`, `mint_erc20/1`, `lock_erc20/1` raise an error when supplied a child key
  - `sign_transaction/2` supports a child key for signing
- Updated `EWallet.AddressTracker`
  - Updated the transaction detection handler to trigger a new process that runs `DepositPoolingGate.on_blockchain_transaction_received/1` so deposit pooling gets triggered and runs independently from the address tracker.
- Updated `EWallet.TransactionTracker`
  - Added `lookup/2` to lookup a tracker's pid by its blockchain transaction hash
  - Added `start/3` that starts a new tracker with the given blockchain transaction hash, registers it with `EWallet.TransactionTrackerRegistry` and dynamically supervises by `EWallet.TransactionTrackerSupervisor`, all in one go.
  - Updated the existing confirmations count handlers to re-retrieve the transaction from database to avoid stale transaction data residing in the GenServer

## Keychain
- Updated `Keychain.Key`
  - Used `uuid` as the primary key instead of `wallet_id` for consistency and avoid confusion of what `wallet_id` means (it meant the address of the `EWalletDB.Wallet`)
  - Added `:wallet_address` to store the `EWalletDB.Wallet`'s address.
- Updated `Keychain.Signature`
  - Added `sign_transaction_hash/6` to support transaction signing with a HD wallet's child key
- Updated `Keychain.Wallet`
  - Added `root_hd_path_public/0` for convenience of retrieving the HD path for public child key derivation
  - Added `root_hd_path_private/0` for convenience of retrieving the HD path for public child key derivation

## Deposit pooling
- Updated `EWallet.BlockchainDepositWalletGate`
  - Updated `do_generate/2` from `case` to `with` control flow for readability & more flexible error handling
  - Added `refresh_balances/2` to refresh balances for the given wallet and list of target tokens.
- Added `EWallet.DepositPoolingGate`
  - Added `move_deposits_to_pooled_funds/1` that takes a `blockchain_identifier` as the only argument, loop through all deposit wallet balances to pool their funds into the primary hot wallet if the criteria are met (current criteria: pool when there's more funds than the gas cost)
  - Added `on_blockchain_transaction_received` that checks if the blockchain transaction is relevant to a deposit wallet. Then creates a new `DepositTransaction` if the funds is incoming to the deposit wallet, or updates an existing `DepositTransaction` by associating it to an `EWalletDB.Transaction` for that blockchain transaction (this will be refactored to `EWalletDB.BlockchainTransaction` in following PRs)
- Added `EWallet.DepositWalletPoolingTracker`
  - Add `start_link/1` and `init/1`, the usual GenServer callbacks
  - The GenServer polls every given interval and simply calls `DepositPoolingGate.move_deposits_to_pooled_funds/1` to do the pooling logic. The GenServer itself does not hold the pooling logic.

## Database schemas
- Updated `EWalletDB.BlockchainDepositWallet`
  - Associates with `EWalletDB.Wallet` with `:wallet_uuid` instead of the address (for consistency that all associations use uuids, also not to confuse which address is which)
  - Removed `:public_key`, not used anywhere + can be derived
  - Renamed `:path_ref` to `:relative_hd_path` for better clarity
- Updated `EWalletDB.Transaction`
  - Change the no longer associated `:from_blockchain_wallet` to a string `: from_blockchain_address` instead
- Updated `EWalletDB.Wallet`
  - Changed `blockchain_deposit_wallets` association to reference by the wallet's uuid instead of address
  - Add `:relative_hd_path` used to build the full HD path
  - Add `get_or_generate_hd_path/1` to get the existing HD wallet or generate a new one
- Added `EWalletDB.BlockchainDepositWalletCachedBalance`
  - Added `all_for_token/2` to retrieve all balances for the given token
  - Added `create_or_update_all/3` to update balances for the given addresses
- Added `EWalletDB.DepositTransaction`
  - Added the usual `get/2`, `get_by/2`, `insert/1`, `update/2`
  - Added `all_unfinalized_by/1` to retrieve all deposit transactions that are present but not finalized on the blockchain. Useful for deducting from the blockchain balance to exclude funds that cannot be taken into calculations.

## Test factory
- Updated `EWalletDB.Factory`
  - Changed default `blockchain_identifier` from `"dumb"` to `"ethereum"` since dumb is set at adapter level and should take `"ethereum"` as the rootchain
  - Added factories for `blockchain_deposit_wallet_cached_balance` and `deposit_transaction`

# Usage

Perform the following steps:

1. Setup an eWallet with `mix seed --sample && mix seed --blockchain`
2. Create a deposit wallet for a wallet via `iex -S mix`, e.g.

    ```elixir
    EWalletDB.Account.all()
    |> Enum.at(-1)
    |> EWalletDB.Account.get_primary_wallet()
    |> EWallet.BlockchainDepositWalletGate.get_or_generate(%{"originator" => 
    %ActivityLogger.System{}})
    ```

3. Deposit some amount of ETH into the above blockchain deposit wallet
4. Wait until the funds is deposited with enough (>= 10) confirmations
5. Manual-trigger the pooling function:

    ```elixir
    EWallet.DepositPoolingGate.move_deposits_to_pooled_funds("ethereum")
    ```

6. You should then see the funds get pooled to the primary hot wallet

# Impact

- No API changes
- New DB schemas
- ⚠️ Modified migrations: Do `mix reset` to drop & remigrate the databases


apps/admin_api/test/support/conn_case.ex
```diff
@@ -147,11 +147,11 @@ defmodule AdminAPI.ConnCase do
 
     config_pid = start_config_server(account, blockchain_wallet)
 
-    {:ok, keychain_hd_wallet_uuid} = Wallet.generate_hd()
+    {:ok, keychain_hd_wallet} = Wallet.generate_hd()
 
     {:ok, _} =
       BlockchainHDWallet.insert(%{
-        keychain_uuid: keychain_hd_wallet_uuid,
+        keychain_id: keychain_hd_wallet.wallet_id,
```
T-Dnzt: The `uuid` was used to avoid leaking any information about the HD Wallet (pkey, public key, address) and keep it all encapsulated in the Keychain app (for security reasons).

apps/eth_blockchain/lib/eth_blockchain/transaction.ex
```diff
@@ -134,6 +134,10 @@ defmodule EthBlockchain.Transaction do
   Returns {:ok, tx_hash, contract_address} if success,
   {:error, code} || {:error, code, message} otherwise
   """
+  def create_contract(%{wallet: _}) do
+    raise ArgumentError, message: "Create a contract with a child key is not supported."
```
T-Dnzt: `"Creating  a contract with a child key is not supported."`

apps/ewallet/lib/ewallet/gates/blockchain/deposit_pooling_gate.ex
```diff
@@ -0,0 +1,145 @@
...
+    # We loop by token because it's likely we pool per specific token than a specific wallet.
+    # E.g. some tokens have more deposits than others, or the hot wallet may run out of one token
+    # more often than others. In this case, loop by token is O(n) while loop by wallet is O(n2).
+    pool_transactions =
+      blockchain_identifier
+      |> Token.all_blockchain()
+      |> Enum.flat_map(fn token ->
+        case token.blockchain_address do
+          ^primary_token_address ->
+            pool_token_deposits(token, hot_wallet, blockchain_identifier, @gas_price, 21_000)
```
T-Dnzt: Don't want to put the amount (`21_000`) as module variable as well? So we can remember to update it in the `TODO`.

apps/ewallet/lib/ewallet/gates/blockchain/deposit_pooling_gate.ex
```diff
@@ -0,0 +1,145 @@
...
+
+          _ ->
+            pool_token_deposits(token, hot_wallet, blockchain_identifier, @gas_price, 90_000)
```
T-Dnzt: Same here.

mederic-p:
A few comments first after initial testing:

apps/ewallet_db/lib/ewallet_db/deposit_transaction.ex
```diff
@@ -0,0 +1,254 @@
...
+    |> validate_blockchain_address(:from_blockchain_address)
+    |> validate_blockchain_address(:to_blockchain_address)
+    |> validate_blockchain_identifier(:blockchain_identifier)
+    |> unique_constraint(:unique_hash_constraint, name: :unique_hash_constraint)
+    |> assoc_constraint(:token)
+    |> assoc_constraint(:from_blockchain_wallet)
+    |> assoc_constraint(:to_blockchain_wallet)
+  end
+
+  defp update_changeset(%DepositTransaction{} = transaction, attrs) do
```
mederic-p: This seems to be unused

unnawut:
TODOs
- [x] Start `DepositWalletTracker` on application start
- [ ] Save `path_ref` when generating `BlockchainDepositWallet`
- [ ] Periodically update `BlockchainDepositWalletBalance` (or keep the balance in sync some other way)
- [ ] Check that a pooling transaction can be completed successfully
- [ ] Transition deposit transactions to failed if it received an error

unnawut:
Note that I removed `deposit_transaction.status` and related blockchain fields completely, and added `deposit_transaction.transaction_uuid` instead. It was too complicated to fix these fields because it was sharing the whole code with `Transaction` but they're not compatible, e.g. 

- `Transaction` requires `local_ledger_uuid` upon confirmation, but `DepositTransaction` does not
- `Transaction` has `from_token` and `to_token`, but `DepositTransaction` simply has `token`
- etc.

In addition to above, the blockchain transactions are now only discovered via `AddressTracker` and the deposit pooling feature simply plugs a one-line `DepositPoolingGate.on_blockchain_transaction_received(transaction)` into the `AddressTracker` to do its own stuff.

Since we're going to extract `transaction`'s blockchain fields into `blockchain_transaction`, it's easier to prepare `deposit_transaction` this way so the only changes needed will be changing `deposit_transaction.transaction_uuid` to `deposit_transaction.blockchain_transaction_uuid`.

All of this should make #1157 and #1158 a lot easier.

unnawut:
Remaining:
- [ ] Pooling only if the deposit wallet proportion to hot wallet reaches a certain threshold
- [ ] Send ETH from hot wallet for pooling if not enough gas in the deposit wallet

unnawut:
@mederic-p @T-Dnzt There's only one nagging random GenServer stop error. Apart from that it should be good to go.

apps/ewallet/test/ewallet/gates/blockchain/deposit_pooling_gate_test.exs
```diff
@@ -0,0 +1,173 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+defmodule EWallet.DepositPoolingGateTest do
+  use EWallet.DBCase, async: false
+  import EWalletDB.Factory
+  alias EthBlockchain.GasHelper
```
mederic-p: Shouldn't alias EthBlockchain here

apps/ewallet/test/ewallet/trackers/address_tracker_test.exs
```diff
@@ -154,21 +173,18 @@ defmodule EWallet.AddressTrackerTest do
...
-                          erc20_address: erc20_token.blockchain_address,
-                          other_address: other_address
-                        }
-                      ]},
-                   node_adapter_name: :dumb,
-                   blockchain_identifier: blockchain_identifier,
-                   stop_once_synced: true
-                 }
+                 node_adapter:
+                   {:dumb_receiver, EthBlockchain.DumbReceivingAdapter,
```
mederic-p: Shouldn't have EthBlockchain here

PR #1161, opened 2019-09-13
Make blockchain addresses case insensitive

### LINKED ISSUE #1160: Support lower and upper case for blockchain addresses

mederic-p:
Issue/Task Number: 1160
Closes #1160 

# Overview

This PR updates the db column of stored blockchain address to be case insensitive so they can be searched given a string ignoring its case.
It also force downcase blockchain addresses before saving them in the DB.


unnawut:
When user sends in a mixed case address, there's some expectation that the address is invalid if it's mistyped/mis-copypasted. We should validate the checksum when it is an external input.

The address can be saved as all lowercase in the database I think, easier for search. I see Etherscan actually does that too, pass the address around as lower case and only encoding back to mixed case on display.

PR #1164, opened 2019-09-23
Validate token blockchain status

### LINKED ISSUE #1148: Prevent blockchain transaction for tokens that are not confirmed

mederic-p:
Issue/Task Number: 1148
Closes #1148 

# Overview

This PR checks for token blockchain status before doing transaction to make sure that the token is confirmed.
Note that until https://github.com/omisego/ewallet/issues/1090 is done, tokens will remain in a `pending` status.

# Changes

- Exclude non confirmed blockchain from `blockchain_wallet.get_balance`
- Prevent `transaction.create` to succeed if the token is not confirmed
- Prevent `blockchain_wallet.deposit_to_childchain` to succeed if the token is not confirmed
- Prepare check when minting blockchain token (future PR)

### DIFF
--- a/apps/admin_api/test/admin_api/v1/controllers/blockchain_wallet_controller_test.exs
+++ b/apps/admin_api/test/admin_api/v1/controllers/blockchain_wallet_controller_test.exs
@@ -66,7 +66,11 @@ defmodule AdminAPI.V1.BlockchainWalletControllerTest do
   describe "/blockchain_wallet.deposit_to_childchain" do
     test_with_auths "deposit to childchain with the given attributes" do
       hot_wallet = BlockchainWallet.get_primary_hot_wallet(@valid_rootchain_identifier)
-      token = insert_confirmed_token("0x0000000000000000000000000000000000000000")
+
+      token =
+        insert(:blockchain_confirmed_token, %{
+          blockchain_address: "0x0000000000000000000000000000000000000000"
+        })
 
       adapter = BlockchainHelper.adapter()
       {:ok, _adapter_pid} = adapter.server().start_link([])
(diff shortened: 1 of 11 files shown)


apps/admin_api/test/admin_api/v1/controllers/blockchain_wallet_controller_test.exs
```diff
@@ -680,4 +578,20 @@ defmodule AdminAPI.V1.BlockchainWalletControllerTest do
       assert response["data"]["code"] == "client:invalid_parameter"
     end
   end
+
+  defp insert_confirmed_token(address, id \\ nil) do
```
unnawut: Can this be moved into `EWalletDB.Factory`?

apps/ewallet/lib/ewallet/gates/mint_gate.ex
```diff
@@ -31,8 +31,12 @@ defmodule EWallet.MintGate do
     mint_token(token, attrs)
   end
 
-  def mint_token(%{blockchain_address: address}, _) when not is_nil(address) do
-    {:error, :invalid_parameter, "A blockchain-enabled token cannot be minted."}
+  def mint_token(%{blockchain_address: address} = token, _) when not is_nil(address) do
+    with true <- Token.blockchain_confirmed?(token) || {:error, :token_is_not_confirmed} do
```
unnawut: I think in this specific code, case is easier to understand.

```elixir
case Token.blockchain_confirmed?(token) do
  true ->
    {:error, :invalid_parameter, "A blockchain-enabled token cannot be minted."}

  false ->
    {:error, :token_is_not_confirmed}
end
```

Which seems to me, we can even simply return the "blockchain-enabled token cannot be minted" error for both cases?

mederic-p: I'll go with the case, this was just to prepare for next PR where we will be able to mint blockchain tokens. Just so wee don't forget to check for token status :)

unnawut: Oh ok I didn't realize it's part of next PR. Makes sense to go with with clauses in that case.

